PR #4, opened 2016-04-29
Contrast 8163 - Updated Contrast SDK to V3

jleo32:
Refactored some of the code to suit V3 API.

Expanded our API calls with additional parameters where needed.

Added Trace search with filter.

Added basic unit tests.


src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -0,0 +1,327 @@
...
+    public HttpURLConnection makeConnection(String url, String method) throws IOException {
+        HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+        connection.setRequestMethod(method);
+        connection.setRequestProperty(RequestConstants.AUTHORIZATION, ContrastSDKUtils.makeAuthorizationToken(user, serviceKey));
+        connection.setRequestProperty(RequestConstants.API_KEY, apiKey);
+        connection.setUseCaches(false);
+        return connection;
+    }
+
+    private static final String GET_REQUEST = "GET";
```
cgtwin: Use /contrast-sdk-java/src/main/java/com/contrastsecurity/http/HttpMethod.java instead of this harcoded strings

cgtwin:
@jleo32 The first endpoint required by any user is GET /ng/profile/organizations, please add it.

src/main/java/com/contrastsecurity/http/UrlBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+    public String getCoverageUrl(String organizationId, String appId) {
+        return String.format("/ng/%s/applications/%s/coverage", organizationId, appId);
+    }
+
+    public String getLibrariesUrl(String organizationId, String appId) {
+        return String.format("/ng/%s/applications/%s/libraries%s", organizationId, appId, buildExpand("manifest", "servers", "cve"));
+    }
+
+    public String getTracesUrl(String organizationId, String appId) {
+        return String.format("/ng/%s/traces/%s/filter/tags/empty-tags/search", organizationId, appId);
```
cgtwin: use /workflow/00001 instead of /tags/empty-tags

src/main/java/com/contrastsecurity/http/UrlBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public String getTracesWithFilterUrl(String organizationalId, String appId, FilterForm form) {
+        return String.format("/ng/%s/traces/%s/filter/tags/empty-tags/search%s", organizationalId, appId, form.toString());
```
cgtwin: use /workflow/00001 instead of /tags/empty-tags

cgtwin:
I'd to see the expand parameters as enums in the SDK, and we should allow the user to decide which expand values want instead of hardcoding them in the URLBuilder class.

jleo32:
@cgtwin I've pushed fixed for everything except the enums. I need to add them to the additional methods but I've removed the hardcoded instance in getLibraries().

PR #15, opened 2017-11-15
CONTRAST-19000 Encode tags when generating URLs

gmoore:
This was my first swing at the problem as I understood it. Let's chat if I'm heading in the wrong direction.

### DIFF
--- a/src/main/java/com/contrastsecurity/http/TraceFilterForm.java
+++ b/src/main/java/com/contrastsecurity/http/TraceFilterForm.java
@@ -4,6 +4,8 @@ import lombok.Getter;
 import lombok.Setter;
 import org.apache.commons.lang.StringUtils;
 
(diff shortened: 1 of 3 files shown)


SteveFeldman:
Two quick questions...

1. Looks like we have a downward dependency that could break a consumer of the plugin. Specifically, the Jenkins plugin pom references version 2.3 of the SDK. Could this break the Jenkins plugin?

Should we look to create a downstream build task for consumers of the SDK?

2. Might be worth writing a test on this?

gmoore:
I assumed we'd bump the Jenkins plugin POM reference and re-release it after this gets approved.

Handled with a test here: https://github.com/Contrast-Security-OSS/contrast-sdk-java/pull/15/files#diff-3179933ae7ba344fb5a2951cc87b49af

PR #16, opened 2018-03-23
Contrast 21352 add proxy support

serge17:
Added proxy support

### DIFF
--- a/src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
+++ b/src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
@@ -55,6 +55,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
+import java.net.Proxy;
 import java.net.URL;
 import java.util.EnumSet;
 import java.util.List;
@@ -71,6 +72,7 @@ public class ContrastSDK {
     private String restApiURL;
     private UrlBuilder urlBuilder;
     private Gson gson;
+    Proxy proxy;
     
     private int connectionTimeout = DEFAULT_CONNECTION_TIMEOUT;
     private int readTimeout = DEFAULT_READ_TIMEOUT;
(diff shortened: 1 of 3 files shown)


src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -552,7 +577,13 @@ public InputStream makeRequest(HttpMethod method, String path) throws IOExceptio
     }
 
     public HttpURLConnection makeConnection(String url, String method) throws IOException {
-        HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+        HttpURLConnection connection;
+        if (this.proxy == null) {
```
gmoore: In the constructor for ContrastSDK without Proxy (https://github.com/Contrast-Security-OSS/contrast-sdk-java/pull/16/files?diff=split#diff-f40b6dbe37b70610f50eaf318983c50dR136), set `self.proxy` to Proxy.NO_PROXY.

Then, you can do-away with this `if` check. The `if` becomes

`connection = (HttpURLConnection) new URL(url).openConnection(this.proxy);`

And since this.proxy is set to NO_PROXY, it'll just pass through. See: https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#openConnection(java.net.Proxy)

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -100,6 +102,29 @@ public ContrastSDK(String user, String serviceKey, String apiKey, String restApi
         this.gson = new Gson();
     }
 
+    /**
+     * Create a ContrastSDK object that will attempt to use the Contrast V3 API
```
gmoore: Change this to 

`Create a ContrastSDK object to use the Contrast V3 API through a Proxy.`

Also change line 139 to

`Create a ContrastSDK object to use the Contrast V3 API`

serge17: Done

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
serge17: Done

PR #26, opened 2019-06-05
Adds /api when necessary

aliaandersen:
If needed, will add /api to a api url that has /Contrast/ or /Contrast


matt-:
Can we remove the .DS_Store files from this?

cixel:
May help to add a `.gitignore` to prevent random files like this from being re-added.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -617,6 +617,6 @@ public void setReadTimeout(final int timeout) {
     private static final int SERVER_ERROR = 500;
 
     private static final String DEFAULT_API_URL = "https://app.contrastsecurity.com/Contrast/api";
-    private static final String LOCALHOST_API_URL = "http://localhost:19080/Contrast/api";
+    private static final String LOCALHOST_API_URL = "http://localhost:19090/Contrast/api";
```
gmoore: I Think this needs to stay at 19080

src/main/java/com/contrastsecurity/utils/ContrastSDKUtils.java
```diff
@@ -35,6 +35,16 @@ public static void validateUrl(String url) throws IllegalArgumentException {
         }
     }
 
+    public static String ensureApi(String url){
+        if(url.endsWith("/Contrast")){
+            url = url + "/api";
+        }
+        else if( url.endsWith("/Contrast/")){
```
gmoore: move the else up a line so it looks like `} else {`

src/test/java/com/contrastsecurity/ContrastSDKTest.java
```diff
@@ -120,5 +119,19 @@ public void negativeTimeoutsAreNotSetTest() throws IOException {
...
+    public void testEnsureApi() throws IOException {
+        final String expectedApi = "http://localhost:19080/Contrast/api";
+        final String ensureUrlOne = ContrastSDKUtils.ensureApi("http://localhost:19080/Contrast/");
+        final String ensureUrlTwo = ContrastSDKUtils.ensureApi("http://localhost:19080/Contrast");
+
+        assertEquals(ensureUrlOne, expectedApi);
+        assertEquals(ensureUrlTwo, expectedApi);
+
+        final String ensureUrlIncorrect = ContrastSDKUtils.ensureApi("http://localhost:19080/");
+        assertNotEquals(ensureUrlIncorrect, expectedApi);
```
gmoore: Change this assertion to assert that the string is unchanged. Our method changes the string if in ends in Contrast, but if it doesn't end in Contrast, then the string should be returned unchanged.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -121,7 +121,10 @@ public ContrastSDK(String user, String serviceKey, String apiKey, String restApi
         this.apiKey = apiKey;
         this.restApiURL = restApiURL;
 
+        this.restApiURL = restApiURL;
```
gmoore: Dupe'd line :)

src/test/java/com/contrastsecurity/ContrastSDKTest.java
```diff
@@ -120,5 +119,20 @@ public void negativeTimeoutsAreNotSetTest() throws IOException {
     	assertNotEquals(connectionTimeout, conn.getConnectTimeout());
     	assertNotEquals(readTimeout, conn.getReadTimeout());
     }
+
+    @Test
+    public void testEnsureApi() throws IOException {
```
gmoore: Good test cleanup. Let's add a few more test cases here: 

* Malformed URL
* Null URL
* Blank URL

But these in separate test methods

PR #27, opened 2019-06-19
Application name endpoint

aliaandersen:
Added the endpoint for /ng/{orgUuid}/applications/name and the request getApplicationsNames to retrieve only the names of the applications. Added to minimize the length of the response when only the name of the application is needed.

### DIFF
--- a/src/main/java/com/contrastsecurity/http/UrlBuilder.java
+++ b/src/main/java/com/contrastsecurity/http/UrlBuilder.java
@@ -39,6 +39,10 @@ public class UrlBuilder {
         return String.format("/ng/%s/applications?%s", organizationId, "base=false");
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/contrastsecurity/http/UrlBuilder.java
```diff
@@ -39,6 +39,10 @@ public String getApplicationsUrl(String organizationId) {
         return String.format("/ng/%s/applications?%s", organizationId, "base=false");
     }
 
+    public String getApplicationsNameUrl(String organizationId) {
+        return String.format("/ng/%s/applications/name?%s", organizationId, "base=false");
```
gmoore: We can clip base=false off of this as it looks like /name doesn't check for base. Don't forget to remove the `%s` off of the first argument.

PR #28, opened 2019-09-15
Customer o

ddooley77:
Hi Gary,

These are changes I made for Optum project last year.

I'm not sure if you want any of these in your SDK.
Maybe you could pick and choose what works.
The GETs are pretty straight fwd.

I had to also do POSTs, and it was a bit cludgey.  It could probably be refactored into common connection stuff.

### DIFF
--- a/.settings/org.eclipse.jdt.core.prefs
+++ b/.settings/org.eclipse.jdt.core.prefs
@@ -1,5 +1,6 @@
 eclipse.preferences.version=1
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.6
-org.eclipse.jdt.core.compiler.compliance=1.6
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
+org.eclipse.jdt.core.compiler.compliance=1.7
 org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
-org.eclipse.jdt.core.compiler.source=1.6
+org.eclipse.jdt.core.compiler.release=disabled
+org.eclipse.jdt.core.compiler.source=1.7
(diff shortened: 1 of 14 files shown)


gmoore:
@ddooley77 Thanks! We'll take a look and get back to you with questions/comments

.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,5 +1,6 @@
 eclipse.preferences.version=1
```
gmoore: Remove this file

csi.properties
```diff
@@ -0,0 +1,4 @@
+username=daviddooley77@gmail.com
```
gmoore: Remove this file. Also rotate these keys if they're still live.

src/test/java/com/contrastsecurity/ddooleyTester.java
```diff
@@ -0,0 +1,256 @@
+package com.contrastsecurity;
```
gmoore: remove this file

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -190,6 +197,314 @@ public Users getOrganizationUsers(String organizationId) throws IOException, Una
...
+     * @return Return HTTP Code.  200 is a success.
+     * @throws UnauthorizedException if the Contrast account failed to authorize
+     * @throws IOException           if there was a communication problem
+     */
+    public int createUser(String organizationId, String userId, String firstName, String lastName, long orgRole, ArrayList<Long> groupIds) throws IOException, UnauthorizedException {
+        HttpURLConnection connection = null;
+        try {
+            String url = restApiURL + this.urlBuilder.createUsersUrl(organizationId);
+            /* Create Connection */
+            connection = (HttpURLConnection) new URL(url).openConnection(this.proxy);
```
gmoore: This code block is duplicated a bunch in these additions. Extract it into something re-usable or make changes to address why we can't re-use makeRequest

PR #29, opened 2019-09-23
Adding Tags to Application model so can be retrieved with getApplication.

ddooley77:
Adding Tags to Application model so can be retrieved with getApplication.


ddooley77:
merging

PR #31, opened 2019-11-04
Adding code to get licensed applications only, as opposed to all apps

sergeib-gmail:
Regions specific changes for only getting licensed application data

### DIFF
--- a/src/main/java/com/contrastsecurity/http/UrlBuilder.java
+++ b/src/main/java/com/contrastsecurity/http/UrlBuilder.java
@@ -39,6 +39,10 @@ public class UrlBuilder {
         return String.format("/ng/%s/applications?%s", organizationId, "base=false");
     }
 
+    public String getLicensedApplicationsUrl(String organizationId) {
+        return String.format("/ng/%s/applications%s", organizationId, "/filter?expand=license,compliance_policy,skip_links,technologies,trace_severity_breakdown,metadata&sort=appName&quickFilter=LICENSED");
+    }
+
     public String getApplicationsNameUrl(String organizationId) {
         return String.format("/ng/%s/applications/name", organizationId);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/contrastsecurity/http/UrlBuilder.java
```diff
@@ -39,6 +39,10 @@ public String getApplicationsUrl(String organizationId) {
         return String.format("/ng/%s/applications?%s", organizationId, "base=false");
     }
 
+    public String getLicensedApplicationsUrl(String organizationId) {
+        return String.format("/ng/%s/applications%s", organizationId, "/filter?expand=license,compliance_policy,skip_links,technologies,trace_severity_breakdown,metadata&sort=appName&quickFilter=LICENSED");
```
gmoore: Can we cut out some(or most) of these expand parameters. Expands really hurt performance.

sergeib-gmail: Yep, took out all the expands, as the script does not seem to need that data. Will push changes shortly.

gmoore:
Hey @sergeib-gmail when I run the following code I get an NPE. How are you using this new method? 

```
            apps = contrastSDK.getLicensedApplications(orgUuid);
            for (Application app : apps.getApplications()) {
                System.out.println(app.getName());
                System.out.println(app.getLicense().getLevel());
            }
```

gmoore:
Oh wait I get it, when you removed the expand param, you removed all of them. Add back only the `license` expand and it should work.

Also feel free to lmk if I'm just missing something.

sergeib-gmail:
Hi @gmoore - I just need a list of all licensed apps,  after that I have this code in my reporting script to get those 5 attributes:

app.getName(), app.getId(), app.getStatus(), app.getArchived(), app.getTags()

I don't need to check for the app.getLicense(), the endpoint is already returning only licensed applications because of the quickFilter=LICENSED in this method in URLBuilder class.

 public String getLicensedApplicationsUrl(String organizationId) {
        return String.format("/ng/%s/applications%s", organizationId, "/filter?sort=appName&quickFilter=LICENSED");
    }

I am happy to put the expand back in though, if you think it's better with it.

gmoore:
Oh interesting, that makes sense. All the same, put the parameter back in. Feels strange to get licensed apps and not the licenses 🚀

PR #33, opened 2019-11-20
Removing debug system.out.println that was left in by mistake.

sergeib-gmail:
Not sure why it's showing line 43 as changed, that expand=license was already in the main repo.  The change was actually to remove a System.out.println() in ContrastSDK.getLicensedApplications() function.  That system.out.println() is not actually in the main repo anyway.  This might be an unneeded push request but I wanted to fix up my fork.


gmoore:
Looks like master is in good shape here.

PR #42, opened 2020-03-26
Add ability to get high vuln count from libs

sethjackson:
Deserialize the `high_vulnerabilities` value in the response.


gmoore:
@sethjackson Thanks! We'll take a look shortly.

PR #46, opened 2020-04-17
Adding functions for getting session metadata, assess tags for applications, get vuln trends, and generate attestation reports

sergeib-gmail:
Various changes for scripts written for Regions Bank as described in the title.

### DIFF
--- a/src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
+++ b/src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
@@ -41,6 +41,7 @@ import com.contrastsecurity.http.TraceFilterType;
 import com.contrastsecurity.http.UrlBuilder;
 import com.contrastsecurity.models.*;
 import com.contrastsecurity.models.dtm.ApplicationCreateRequest;
+import com.contrastsecurity.models.dtm.AttestationCreateRequest;
 import com.contrastsecurity.utils.ContrastSDKUtils;
 import com.contrastsecurity.utils.MetadataDeserializer;
 import com.google.gson.Gson;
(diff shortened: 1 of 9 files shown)


src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -588,14 +610,100 @@ public Traces getTraces(String organizationId, String appId, TraceFilterForm for
...
+
+    /**
+     * Generate an attestation report for the application whose ID is passed in.
+     * @param organizationId the ID of the organization
+     * @param appId          the ID of the application
+     * @param request
+     * @throws IOException
+     * @throws UnauthorizedException
+     */
+    public void generateAttestationReport(String organizationId, String appId, AttestationCreateRequest request)
```
gmoore, line 698: @sergeib-gmail I'm not all that familiar with this feature. How long do these usually take to generate? Seconds? Minutes?

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -188,6 +189,27 @@ public VulnerabilityTrend getYearlyVulnTrend(String organizationId) throws IOExc
         }
     }
 
+    /**
+     * Get all Vulnerability Trend for an Organizations.
```
gmoore, line 193: Can you update this comment to reflect how this method is different from the one above?

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
sergeib-gmail: It depends on the # of vulns/libraries in the application.  It's usually within seconds, but can be minutes for large number of findings.  I actually am not using this function in any script that was sent to a customer, was just testing it as part of a proof of concept to see if an attestation report can be generated via API.  I can take it out/leave it in, whichever you prefer.

PR #54, opened 2020-08-07
Addded getYearlyVulnTrendForApplication function 

mprashant24:
Addded getYearlyVulnTrendForApplication function which is used by Customer reporting, it got missed to merged earlier.


mprashant24:
@gmoore could you please merge the above changes which was done for customer reporting. Customer wants to have the source code of the reporting tool we build. We want to have all the dependency in place before we share the code with customer.

PR #58, opened 2020-11-18
Add start date to security check

aliaandersen:
Adds job start time to security check


pom.xml
```diff
@@ -10,7 +10,7 @@
 
     <groupId>com.contrastsecurity</groupId>
     <artifactId>contrast-sdk-java</artifactId>
-    <version>2.16-SNAPSHOT</version>
```
bryanateoan, line 13: is this intentional?

bryanateoan (changes requested):
revert version change

aliaandersen:
The large version change was not having merged in master after Gary's branches. I synced the branch.

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -60,4 +68,9 @@ public SecurityCheckForm(String applicationName, AgentType agentLanguage) {
         this.applicationName = applicationName;
         this.agentLanguage = agentLanguage;
     }
+
+    public SecurityCheckForm(String applicationId, Long jobStartTime) {
```
bryanateoan, line 72: no need for constructor if jobStartTime isn't going to be a required argument.

bryanateoan:
> The large version change was not having merged in master after Gary's branches. I synced the branch.

🤔 seems I can't reply to your comment. The change is still showing as a diff.

aliaandersen:
Should be 2.19->2.2

bryanateoan:
The release plugin usually changes the version.

aliaandersen:
Oh yep, forgot about that for this.

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
bryanateoan, line 72: If you want job_start_time to be required. Can you mark the old constructors as deprecated and add a constructor for (name,language,jobStartTime) as well as we can remove the old constructors later.

bryanateoan: If you do not want job_start_time to be a required parameter, make a note in the javadocs comment that it is needed for some job outcome policy configurations. If job_start_time isn't going to be required, then we don't need this constructor.

bryanateoan (changes requested):

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
aliaandersen: What other usage are there for the sdk methods? I made that constructor so it could be nice and clean with just applicationId and jobStartTime, since it can be sent by jenkins every time. Is there a common use case for customers using the sdk method for security checks without being connected to jenkins already? Then I could see a stronger need to leave it optional.

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -52,6 +52,14 @@
...
 
+    /**
+     * The job start time of the jenkins job
+     * @param jobStartTime New value for job start time.
+     * @return the start time of the jenkins job
+     */
+    @SerializedName("job_start_time")
+    private Long jobStartTime;
+
     public SecurityCheckForm(String applicationId) {
```
gmoore, line 63:
```
public SecurityCheckForm(String applicationId) { return SecurityCheckForm(applicationId, null); }
public SecurityCheckForm(String applicationId, HashMap<String, String> options)
public SecurityCheckForm(String applicationName, AgentType agentLanguage)
public SecurityCheckForm(String applicationName, AgentType agentLanguage, HashMap<String, String> options)
```

If there's a smarter way to do this with varargs then give that a shot.

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -52,12 +49,31 @@
...
-    public SecurityCheckForm(String applicationId) {
+    /**
+     * The job start time of the jenkins job
+     * @param jobStartTime New value for job start time.
+     * @return the start time of the jenkins job
+     */
+    @SerializedName("job_start_time")
+    private Long jobStartTime;
+
+    public static SecurityCheckForm securityCheckForm(String applicationId) {
```
bryanateoan, line 60: why are these constructors are static?

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -52,12 +49,31 @@
...
+    }
+
+    public SecurityCheckForm(String applicationId, HashMap<String, Object> options) {
         this.applicationId = applicationId;
+        this.jobStartTime = Long.valueOf(options.get("jobStartTime").toString());
     }
 
-    public SecurityCheckForm(String applicationName, AgentType agentLanguage) {
-        this.applicationName = applicationName;
+    public static SecurityCheckForm securityCheckForm(String applicationId, AgentType agentLanguage) {
```
bryanateoan, line 69: same here

bryanateoan: This should be applicationName

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -52,12 +49,31 @@
...
+        return new SecurityCheckForm(applicationId, agentLanguage, null);
+    }
+
+    public SecurityCheckForm(String applicationId, AgentType agentLanguage,  HashMap<String, Object> options) {
```
bryanateoan, line 73: This should be applicationName

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
bryanateoan: If you want to make these static factories instead, we should depricate the old constructors and add 

```
public static SecurityCheckForm newSecurityCheckFormWithId(String applicationId);
public static SecurityCheckForm newSecurityCheckFormWithNameAndLanguage(String applicationName, AgentType agentLanguage);
```
there is probably a better naming scheme...

bryanateoan (changes requested):

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
bryanateoan: lol, I read it wrong, these are constructors.

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
```diff
@@ -52,12 +49,31 @@
     @SerializedName("origin")
     private String origin = "OTHER";
 
-    public SecurityCheckForm(String applicationId) {
+    /**
+     * The job start time of the jenkins job
```
bryanateoan, line 53: Can you add a sentence here that some job outcome policy configurations need the jobStartTime?

src/main/java/com/contrastsecurity/http/SecurityCheckForm.java
gmoore, line 60: public SecurityCheckForm(String applicationId)
public SecurityCheckForm(String applicationId, AgentType agentLanguage)

Add a set method for jobStartTime

gmoore (changes requested):

gmoore (approved):

bryanateoan (approved):

PR #63, opened 2021-04-15
Int 515 telemetry constructor

aliaandersen:
Added integrationName 

- thought about ideName, but since this constructor can be used across more than just the ide's, I opted for something more generic. 

Added a string for version


bryanateoan (approved):
LGTM

src/main/java/com/contrastsecurity/http/RequestConstants.java
```diff
@@ -3,6 +3,8 @@
 public class RequestConstants {
     public static final String AUTHORIZATION = "Authorization";
     public static final String API_KEY = "API-Key";
+    public static final String X_TELEMETRY_INTEGRATION_NAME = "X_TELEMETRY_INTEGRATION_NAME";
+    public static final String X_TELEMETRY_VERSION = "X_TELEMETRY_VERSION";
```
bryanateoan, line 7: Lets make this `X_TELEMETRY_INTEGRATION_VERSION` to be consistent. or make the other `X_TELEMETRY_NAME`

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -66,16 +67,21 @@
...
     private Gson gson;
     Proxy proxy;
-    
+    private String integrationName;
+    private String version;
+
     private int connectionTimeout = DEFAULT_CONNECTION_TIMEOUT;
     private int readTimeout = DEFAULT_READ_TIMEOUT;
 
+    @Getter
```
bryanateoan, line 76: This is probably fine for pragmatism, but when we are using the sdk, we should try to avoid treating the builder like a repository/holder of values.

bryanateoan (approved):

src/main/java/com/contrastsecurity/http/RequestConstants.java
bryanateoan: recreating a review, forgot to mark for request changes.

bryanateoan (changes requested):
Make the X_TELEMETRY consistent with each other.

bryanateoan (approved):

PR #66, opened 2021-05-10
Int 520 removed extended sdk from intellij

aliaandersen:
This adds the necessary headers to the makeConnection method to handle the incoming telemetry values. 

This adds the necessary files to remove the extendedSDK from intellij. 

Bumps java version to 1.8 and updates README accordingly.

Bumps major version for the java change and removes depreciated constructors.


src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -151,7 +160,7 @@ public ContrastSDK(String user, String serviceKey, String apiKey, String restApi
         ContrastSDKUtils.validateUrl(this.restApiURL);
         this.restApiURL = ContrastSDKUtils.ensureApi(this.restApiURL);
         this.urlBuilder = UrlBuilder.getInstance();
-        this.gson = new GsonBuilder()
+        this.gson = new GsonBuilder().excludeFieldsWithoutExposeAnnotation()
```
bryanateoan, line 163: I think this breaks serialized objects that were never tagged with expose like JobOutcomePolicyListResponse

bryanateoan: Maybe we should us transient instead of expose annotation? what do you think?

bryanateoan (changes requested):
Requested some changes and discussion questions.

.java-version
```diff
@@ -1 +0,0 @@
-openjdk64-11.0.2
```
bryanateoan, line 1: Is this expected?

src/main/java/com/contrastsecurity/models/Tags.java
```diff
@@ -0,0 +1,68 @@
+package com.contrastsecurity.models;
+
+import com.google.gson.annotations.Expose;
```
bryanateoan, line 3: Lets remove if no longer needed.

pom.xml
```diff
@@ -107,14 +107,6 @@
             </testResource>
         </testResources>
         <plugins>
-            <plugin>
-                <artifactId>maven-compiler-plugin</artifactId>
```
bryanateoan, line 111: Please update the README and make note in the release notes.

bryanateoan (changes requested):

.java-version
aliaandersen: Nope, I'll add to gitignore.

pom.xml
```diff
@@ -10,7 +10,7 @@
 
     <groupId>com.contrastsecurity</groupId>
     <artifactId>contrast-sdk-java</artifactId>
-    <version>2.23-SNAPSHOT</version>
+    <version>3.0</version>
```
bryanateoan, line 13: Keep it 3.0-SNAPSHOT. The release process will release 3.0 and move the development version to 3.1-SNAPSHOT.

src/main/java/com/contrastsecurity/http/UrlBuilder.java
```diff
@@ -102,9 +102,41 @@ public String getTracesByApplicationUrl(String organizationId, String appId, Tra
...
+        return String.format("/ng/%s/traces/%s/events/summary", organizationId, traceId);
+    }
+
+    public String getEventDetails(String organizationId, String traceId, String eventId){
+        return String.format("/ng/%s/traces/%s/events/%s/details", organizationId, traceId, eventId);
+    }
+
     public String getNotesByApplicationUrl(String organizationId, String appId, String traceId, TraceFilterForm form) throws UnsupportedEncodingException {
         String formString = form == null ? "" : form.toQuery();
-        return String.format("/ng/%s/applications/%s/traces/%s/notes?expand=skip_links", organizationId, appId, traceId, formString);
```
bryanateoan, line 107: Why did you remove the expand=skip_links?

bryanateoan: is that no longer needed?

bryanateoan (changes requested):
* Change the version to 3.0-SNAPSHOT
* Question about getNotesByApplicationURL

Rest looks good.

pom.xml
aliaandersen: Shoot, got to excited moving to v3

src/main/java/com/contrastsecurity/http/UrlBuilder.java
aliaandersen: Yeah, it wasn't necessary for this endpoint

aliaandersen: This endpoint, as far as I know, doesn't return links, so it shouldn't need the expand param

bryanateoan (approved):

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -124,74 +133,19 @@ public ContrastSDK build() {
         }
     }
 
-    /**
-     * Use ContrastSDK.Builder
-     */
-    @Deprecated
     public ContrastSDK() {
```
bryanateoan, line 136: lets make this protected with a visible for testing comment.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -64,11 +72,12 @@
     private String apiKey;
     private String serviceKey;
     private String user;
-    private String restApiURL;
+    @Getter
+    private static String restApiURL;
```
bryanateoan, line 76: Why is this static?

bryanateoan (changes requested):
just a couple of items, rest looks good.

bryanateoan (approved):

PR #69, opened 2021-06-14
👷 JAVA-3134 Replace Travis CI with GitHub Actions

gilday:
* Introduces Maven Wrapper to pin Maven at version 3.8.1
* Updates README to include Maven Wrapper
* Updates README to lead users to Maven Central instead of building the project locally
* Replaces Travis CI with GitHub Actions

PR #70, opened 2021-06-14
🎨 JAVA-3133 Automated Formatting

gilday:
Configures Google Java Format using Spotless Maven plugin

* Google Java Format for Java code
* General newline configuration for all source
* Run format check on JDK 11 in CI

By its nature, this is a noisy PR. The interesting changes are isolated to

* README.md
* pom.xml
* .github/workflows/build.yml

Use the "Ignore Whitespace Changes" setting to see the important changes in those files


gilday:
Keeping in draft until #69 is merged

pom.xml
```diff
@@ -1,284 +1,329 @@
...
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>8</source>
+          <target>8</target>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
```
bryanateoan, line 213: Is spotless better than fmt? I think [fmt](https://github.com/coveooss/fmt-maven-plugin) works on java 8.

bryanateoan (changes requested):
question about spotless vs fmt

pom.xml
gilday: Both tools require Java 11:

```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.11:check (default) on project contrast-sdk-java: Execution default of goal com.coveo:fmt-maven-plugin:2.11:check failed: An API incompatibility was encountered while executing com.coveo:fmt-maven-plugin:2.11:check: java.lang.UnsupportedClassVersionError: com/google/googlejavaformat/java/FormatterException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

A [note in the fmt readme](https://github.com/coveooss/fmt-maven-plugin#using-with-java-8) tells us that the underlying library Google Java Format has migrated to Java 11 starting in version Google Java Format version 1.8.

I am confused by this, because [that repository appears to be configured to target Java 1.8](https://github.com/google/google-java-format/blob/1c45fa55b1a483314f8cb52bd0a0b21838c2bc32/pom.xml#L97), but I can't argue with the bytecode version in the jar 🤷 

As for "is spotless better than fmt", I don't know. Spotless is more flexible, but that could be good or bad. I chose Spotless because it has plugins for both Maven and Gradle, so I like that we can be consistent across our projects instead of using Spotless for Gradle and fmt for Maven.

bryanateoan: ts uses 2.9 which works with 1.8

gilday: It works with 1.8 because it uses an older version of Google Java Format. You can achieve the same effect with the latest Spotless plugin version by configuring it to use Google Java Format 1.7.

However, I don't think we want to do that, because it invites issues with developers' editors. For example, if you have the latest [IntelliJ Google Java Format plugin](https://plugins.jetbrains.com/plugin/8527-google-java-format) installed, then it will use Google Java Format 1.9 which will apply different formatting. Consequently, your IDE formatting and the build's formatting will be out of sync and they will fight each other.

bryanateoan: ah ok

bryanateoan (approved):

pom.xml
gilday: For what it's worth, we're running into this problem because we're using a build matrix to test the project across different versions of Java. When you do that, your entire Maven build must be compatible with each JDK version you need to test with (vs only your code).

Instead of using a build matrix, we could try to migrate the build to [verify the project using Maven Toolchains](https://johnathangilday.com/blog/maven-toolchain-tests/). My take is that this formatting issue isn't enough of a problem to migrate from the relatively simple matrix build to the relatively more complex Maven Toolchains, so we deal with it for now and we can always revisit 🙂

BrianPhillips2020 (approved):

aliaandersen (approved):

stiemannkj1 (approved):

PR #72, opened 2021-06-22
Add V2 Vulnerability Filtering API, fix call to get session metadata

atzawada:
This change implements part of the V2 Application Vulnerability Filtering API, specifically this endpoint:
```
POST  /ng/{orgUuid}/traces/{appId}/filter
```

The other change is to the way the URL is generated for getting the session metadata for an application. Previously the URL would not be generated correctly if the `TraceFilterForm` had no filters or was null.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
bryanateoan, line 1: Please remove this file.

We started adding any IDE specific files into `.gitignore`

src/main/java/com/contrastsecurity/models/TraceFilterBody.java
```diff
@@ -0,0 +1,30 @@
+package com.contrastsecurity.models;
```
bryanateoan, line 1: Any reason we can't reuse `TraceFilterForm` and just add the missing fields?

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -761,6 +761,35 @@ public Traces getTraces(String organizationId, String appId, TraceFilterForm for
     }
   }
 
+  /**
```
bryanateoan, line 764: If we reuse the `TraceFilterForm` we can also just update the existing `getTraces` method to use V2 vulnerability filtering. Both methods are getting a list of vulnerabilities based on a filter.

bryanateoan (changes requested):
Hi Andy, thank you very much for the contribution.

I have a few request and questions.
1. Remove vscode settings file
2. Questions about re-using TraceFilterForm

atzawada:
Hi Brian, is there any reason to maintain the fields on the old v2 TraceFilterForm? Are they used to query in a similar fashion anywhere else in the v3 API?

bryanateoan:
TraceFilterForm is used by `getSessionMetadataForApplicationUrl` and `getTracesByOrganization`. However, similar to getTraces, it is also moving to having its parameters in the body. The `TraceMetadataFilter` can probably be encoded similar to how `appVersionTags` is encoded for the toQuery method.

bryanateoan:
@atzawada, I'm okay with adding a new getTraces method. Can you mark the existing getTraces that uses the TraceFilterForm as `@Deprecated`.

bryanateoan (approved):
Thank you very much for the contribution!

PR #73, opened 2021-06-29
Spotless check for publish

aliaandersen:
This updates the jdk version to 11 for the publish step so that we can run the spotless test in the publish workflow


bryanateoan (approved):
lgtm

PR #78, opened 2021-07-16
Use java.util.Base64 instead of commons-codec

sethjackson:
Use standard library instead of `commons-codec` which drops a dependency.
Since the library is supported on JDK 8 or newer this should be ok to do.


bryanateoan (approved):
Change looks good. Thank you very much for the contribution!

PR #95, opened 2021-08-09
Added Query to facade the change for naming FilterForms to Queries

bryanateoan:
# Background
We are moving to naming FilterForm as Query.

# Purpose
This adds a Query class so that new Query will extend a Query class instead of a FilterForm class.

# What changed
1. Marked FilterForm as deprecated
2. Added Query to act as a facade for new Queries. When FilterForm is removed in the next major version, it will move to Query.
3. Added QueryException so that we can throw exceptions during the toQuery method instead of swallowing and returning a string.

### DIFF
--- a/src/main/java/com/contrastsecurity/http/Query.java
+++ b/src/main/java/com/contrastsecurity/http/Query.java
@@ -1,13 +1,16 @@
 package com.contrastsecurity.http;
 
+import com.contrastsecurity.exceptions.QueryException;
+
 public class Query extends FilterForm {
 
   /**
    * Returns the query into url parameters
    *
    * @return the query into url parameters
+   * @throws QueryException when an exception occurs converting to url parameters
    */
-  public String toQuery() {
+  public String toQuery() throws QueryException {
     return super.toString();
   }
 }
--- a/src/main/java/com/contrastsecurity/http/Query.java
+++ b/src/main/java/com/contrastsecurity/http/Query.java
@@ -1,8 +1,11 @@
 package com.contrastsecurity.http;
 
 import com.contrastsecurity.exceptions.QueryException;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.EnumSet;
+import java.util.List;
+import org.apache.commons.lang.StringUtils;
 
 public class Query<T extends Query<T>> {
   private EnumSet<?> expand;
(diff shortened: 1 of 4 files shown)


src/main/java/com/contrastsecurity/http/Query.java
```diff
@@ -1,9 +1,10 @@
 package com.contrastsecurity.http;
 
-public class Query extends FilterForm{
+public class Query extends FilterForm {
```
tylernhoward, line 5: If at some point we intend to completely remove `FilterForm` we would have to duplicate the methods in that class. We could do that now and have `Query` _not_ extend `FilterForm` and rather just have the same methods/enums.

Another option is that since `FilterForm` only contains two enums we could reverse the organization a bit and use interfaces like so:
```suggestion
public interface Query  {
    **
       * Returns the query into url parameters
       *
       * @return the query into url parameters
       */
      public String toQuery();
}

public enum LibrariesExpandQuery implements Query {
    VULNS;
    
    @Override
    public String toQuery() {
      return name().toLowerCase();
    }
```
This would have the same duplication of code but enforce that method name be used. Then again im not super familiar with what these enums are :)

tylernhoward:
Yo what is a draft pr - this is dope.

src/main/java/com/contrastsecurity/http/Query.java
bryanateoan: I like it 👍

bryanateoan:
We've used draft PRs for prs that are not yet ready to be merged.
"But why open a PR if it's not ready to be merged?" you may ask. we usually do it for 2 reasons.
1. We want some eyes to take a look so that we can gather opinions but are not yet ready to merge.
2. We want to make changes to an open pr, so we convert it to a draft.

tylernhoward (approved):

src/main/java/com/contrastsecurity/http/Query.java
```diff
@@ -0,0 +1,135 @@
+package com.contrastsecurity.http;
+
+import com.contrastsecurity.exceptions.QueryException;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.EnumSet;
+import java.util.List;
+import org.apache.commons.lang.StringUtils;
+
+public class Query<T extends Query<T>> {
```
tylernhoward, line 10: What are all these new changes?

bryanateoan:
Closing this pull request. Naming strategies are still in draft.

PR #96, opened 2021-08-17
INT-773 Update JavaSDK Telemetry Headers

arturoespinoza15:
**Description**

Currently we are sending the IDE telemetry header values, those header values are custom and we need to use User-Agent header to send those values to TS.

**Changes:**

-  method to build User-Agent header with integration name and version.
-  test added for `makeConnection` method to check if the user agent header is built


pom.xml
```diff
@@ -89,6 +95,17 @@
...
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+      <version>5.2.9.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>org.testng</groupId>
+      <artifactId>testng</artifactId>
+      <version>6.14.3</version>
+      <scope>test</scope>
+    </dependency>
```
bryanateoan, line 108: Why is spring included?

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1830,13 +1831,29 @@ public HttpURLConnection makeConnection(String url, String method) throws IOExce
       connection.setRequestProperty(RequestConstants.TELEMETRY_INTEGRATION_VERSION, version);
     }
 
+    connection.setRequestProperty(HttpHeaders.USER_AGENT, buildUserAgentString().toString());
```
bryanateoan, line 1834: Are you able to append to the header instead of setting it so that we can keep the other User-Agent information or is that automatic?

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1857,6 +1874,16 @@ public void setReadTimeout(final int timeout) {
...
+  /** Set a integrationName for all SDK requests. This method is only for test don't use it */
+  public void setIntegrationName(final IntegrationName integrationName) {
+    this.integrationName = integrationName;
+  }
+
+  /** Set a version for all SDK requests. This method is only for test don't use it */
+  public void setVersion(final String version) {
+    this.version = version;
+  }
+
```
bryanateoan, line 1886: Make these package private if they are only for testing.

src/test/java/com/contrastsecurity/ContrastSDKTest.java
```diff
@@ -373,4 +378,54 @@ public void testGetTraceMetadataFilters() {
...
+  public void make_connection_validate_user_agent_header(
+      final String testCase,
+      final IntegrationName integrationName,
+      final String version,
+      final String expectedResult)
+      throws IOException {
+
+    final ContrastSDK sdk = buildSdk(integrationName, version);
+
+    URLConnection conn = sdk.makeConnection("https://www.google.com", HttpMethod.GET.toString());
```
bryanateoan, line 409: is this making an actual connection to google?

pom.xml
bryanateoan: use junit instead of testng

src/test/java/com/contrastsecurity/ContrastSDKTest.java
bryanateoan: feels wrong to make an actual connection for unit tests.

bryanateoan (changes requested):
Some Questions and change requests

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -61,6 +61,7 @@
 import java.util.Set;
 import lombok.Getter;
 import org.apache.commons.io.IOUtils;
+import org.springframework.http.HttpHeaders;
```
gilday, line 64: Spring is far too big of a dependency to include for an HTTP header constant. Let's just define our own `USER_AGENT` constant if we'd like to use a constant (though I expect the string will only appear in exactly one place is a well-known standard, so I'd recommend foregoing the constant altogether).

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
gilday: Unfortunately, I don't think there's a way to get the default user-agent header that java builds so that we can prepend to it.

I tried this, but it returns `null`

```
connection.getRequestProperty("User-Agent");
```

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1830,13 +1831,29 @@ public HttpURLConnection makeConnection(String url, String method) throws IOExce
...
+    connection.setRequestProperty(HttpHeaders.USER_AGENT, buildUserAgentString().toString());
+
     connection.setUseCaches(false);
     if (connectionTimeout > DEFAULT_CONNECTION_TIMEOUT)
       connection.setConnectTimeout(connectionTimeout);
     if (readTimeout > DEFAULT_READ_TIMEOUT) connection.setReadTimeout(readTimeout);
     return connection;
   }
 
+  private StringBuilder buildUserAgentString() {
```
gilday, line 1843: This builds a string that only contains the integration name and version. We need to also include:

* comments
* the Contrast Java SDK name and version
* the JRE platform name and version

For example:

```
User-Agent: contrast-maven-plugin/2.12 (Apache Maven 3.8.1) contrast-sdk-java/2.22 Java/11.0.6
```

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
gilday: I recommend making these fields `final` and rebuilding a ContrastSDK to change them instead of introducing this mutability just for testing.

src/test/java/com/contrastsecurity/ContrastSDKTest.java
gilday: This _shouldn't_ actually make an outgoing connection, because we haven't called any methods on the `HttpURLConnection` that would cause the connection to start. However, we should not connect to google.com for testing. If we want to be sure that we're not sending an outgoing connection in this test, then we should send the connection to a server we know will fail e.g. `does-not-exist.contrastsecurity.com`.

If I'm wrong about this, or if we or the JDK changes `makeConnection` such that it does cause an outgoing request, then we should reconsider the design of this test. We should instead use a mock server that is capable of receiving the request and verifying that we sent the User-Agent header.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1857,6 +1874,16 @@ public void setReadTimeout(final int timeout) {
     this.readTimeout = timeout;
   }
 
+  /** Set a integrationName for all SDK requests. This method is only for test don't use it */
+  public void setIntegrationName(final IntegrationName integrationName) {
```
gilday, line 1878: I think the `IntegrationName` enum is a mistake. 

1. The enum tightly couples the SDK to each integration we've implemented. The SDK should be unaware of downstream projects.
1. The enum leaves no room for users to add their own value. It's not inconceivable that we'd want users to have an opportunity to tell us what they're building with the SDK.

I recommend eliminating the `enum` and accepting a simple `String` instead.

gilday (changes requested):

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1830,13 +1831,29 @@ public HttpURLConnection makeConnection(String url, String method) throws IOExce
...
     if (connectionTimeout > DEFAULT_CONNECTION_TIMEOUT)
       connection.setConnectTimeout(connectionTimeout);
     if (readTimeout > DEFAULT_READ_TIMEOUT) connection.setReadTimeout(readTimeout);
     return connection;
   }
 
+  private StringBuilder buildUserAgentString() {
+    final StringBuilder userAgent = new StringBuilder();
+
+    if (integrationName != null) {
```
gilday, line 1846: Can we rename `integrationName` and `integrationVersion` to `componentName` and `componentVersion`?

"integration" is something that only Contrast cares about, but this SDK is a general library that users can build things with too. "Component" comes from the language describing the User-Agent header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```diff
@@ -1830,13 +1831,29 @@ public HttpURLConnection makeConnection(String url, String method) throws IOExce
       connection.setRequestProperty(RequestConstants.TELEMETRY_INTEGRATION_VERSION, version);
```
gilday, line 1831: Let's remove these proprietary headers. The User-Agent is intended to replace these headers.

src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
gilday: I used the word "product" instead of "component", because that is consistent with the HTTP specification.

gilday (approved):

bryanateoan (approved):

rduran0 (approved):

aliaandersen (approved):

PR #97, opened 2021-08-17
❇️ JAVA-3298 Contrast Scan Support

gilday:
Adds support for the Contrast Scan API.

* Adds Pact framework for generating Pacts for the Contrast Scan API. Out of scope: publishing pacts to our Pact Broker (to be done in a subsequent PR). 
* Removes legacy, defunct screener test.
* Aligns tests to package structure.
* Scan package exposes high-level API, low-level API is package private.
* Introduces `internal` package for public classes that we use internally but do not consider part of our public API.
* Introduces `URIBuilder` to help us build URIs that account for URL encoding.
* Introduces new exception type hierarchy. Developers should use `HttpResponseException.fromConnection` to capture all of the details about a failed response.

I added an [addendum to our internal design document](https://contrast.atlassian.net/wiki/spaces/JAVA/pages/2483519785/High-Level+Contrast+Java+SDK#Addendum:-Implementation-Details) to help explain to reviewers some of the choices made in this implementation.


src/main/java/com/contrastsecurity/sdk/scan/ScanException.java
```diff
@@ -0,0 +1,23 @@
+package com.contrastsecurity.sdk.scan;
+
+import java.util.Objects;
+
+/** Indicates a condition that prevents Contrast Scan from completing an analysis */
+public class ScanException extends RuntimeException {
```
bryanateoan, line 6: Why does this not extend ContrastException instead?

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
```diff
@@ -0,0 +1,52 @@
+package com.contrastsecurity.sdk.scan;
+
+import com.google.auto.value.AutoValue;
+import java.time.Instant;
+
+/** Value type that describes the code artifact structure returned by the Contrast API. */
+@AutoValue
+abstract class CodeArtifactInner {
```
bryanateoan, line 8: Opinion question, for these `Inner` classes. What do you think about using an `Api` prefix instead, `ApiCodeArtifact`?

pom.xml
```diff
@@ -36,12 +36,11 @@
...
   <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <localTeamServerUrl>http://localhost:19080/Contrast/api</localTeamServerUrl>
-    <versions.maven-source-plugin>3.0.1</versions.maven-source-plugin>
-    <versions.maven-javadoc-plugin>3.0.0</versions.maven-javadoc-plugin>
-    <versions.maven-gpg-plugin>1.6</versions.maven-gpg-plugin>
-    <versions.nexus-staging-maven-plugin>1.6.8</versions.nexus-staging-maven-plugin>
     <versions.junit-jupiter>5.7.1</versions.junit-jupiter>
+    <versions.auto-value>1.8.2</versions.auto-value>
+    <versions.lombok>1.18.18</versions.lombok>
```
tylernhoward, line 43: assuming lombok is used in the same capacity as AutoValue in the old code and form objects? We really should prefer autovalue imo, so this is a good start - just weird we have both for the time being

src/main/java/com/contrastsecurity/exceptions/ContrastException.java
```diff
@@ -0,0 +1,15 @@
+package com.contrastsecurity.exceptions;
+
+/** Generic {@link RuntimeException} thrown by Contrast code. */
+public class ContrastException extends RuntimeException {
```
tylernhoward, line 4: We have this same wrapped RuntimeException in the java-agent codebase. Is this a common pattern? It makes more sense for the agent in some ways because it makes it apparent for the users that the issue is one with Contrast, but for the sdk I feel like `HttpResponseException` could just extend `RuntimeException` given the scope of this project. Do other SDKs wrap their exceptions in a way like this?

bryanateoan (approved):
everything lgtm, just a couple of questions?

src/main/java/com/contrastsecurity/exceptions/HttpResponseException.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  /**
+   * Reads the response body in its entirety. Error responses should always be relatively small, so
+   * they should fit into memory.
+   *
+   * @return body as a string, or {@code null}
+   */
+  private static String readBody(final HttpURLConnection connection, final int code)
```
tylernhoward, line 61: This kinda sucks. You think TeamServer will ever use proper response codes?

aliaandersen (approved):

pom.xml
gilday: Right, it's still used by the old code for more than just making value types. We'll leave it in while we transition. I got them to place nicely together 🙂

src/main/java/com/contrastsecurity/exceptions/ContrastException.java
gilday: > Do other SDKs wrap their exceptions in a way like this?

yes. examples:

* https://azure.github.io/azure-sdk/java_introduction.html
* https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/exception/SdkException.html
* https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.13.0-rc1/com/fasterxml/jackson/core/JacksonException.html

src/main/java/com/contrastsecurity/exceptions/HttpResponseException.java
gilday: I think you're referring to the legacy TeamServer code that returns 200 and an error object in the response body. This code is actually not related to that. In that case, the code making the request would have already had to read the body to determine if an exception needs to be thrown.

This method is used when the code making the request has determined that it doesn't like the status code returned by the API, and it wants to throw a meaningful exception. This code reads debug information from the body of the response to help you understand why the request failed. For example TeamServer might return:

```
HTTP/2 400
date: Fri, 20 Aug 2021 15:53:31 GMT
content-type: application/json
content-length: 46

{"error":{"message":"Missing application ID"}}
```

If the `HttpResponseException` _only_ had the status code and some message like "failed to read vulnerabilities", the developer reading the exception would have no way to know the error message returned in the response body. This code makes sure the developer gets that information so they don't have to do anything painful like put Wireshark between the SDK and TeamServer to debug.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
gilday: I'm not in love with the `Inner` suffix either, but it is [consistent with the types used by the Azure Java SDK](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/resourcemanager/azure-resourcemanager-compute/src/main/java/com/azure/resourcemanager/compute/fluent/models), which you'll recall is the example we're referencing for this design. Like we say in the Java Team style guide, we like to avoid bike-shedding opinions by adopting some existing standard, so I'm cool with just doing what they do, even if I don't love it.

I do have opinions on suffix vs prefix. I strongly prefer a suffix in this case, because in an alphabetical listing of classes, we want all the classes related to a resource to be grouped together e.g.

```
Project.java
ProjectClient.java
ProjectClientImpl.java
ProjectCreate.java
ProjectImpl.java
ProjectInner.java
Projects.java
ProjectsImpl.java
```

If you use a prefix, then you'll group classes together by their layer archetype instead of by resource. In this resource-centric design, we want to package by feature and group by resource.

src/main/java/com/contrastsecurity/sdk/scan/ScanException.java
gilday: It should. Copy / paste error. I'll fix it

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
```diff
@@ -0,0 +1,58 @@
...
+import com.contrastsecurity.exceptions.ResourceNotFoundException;
+import com.contrastsecurity.exceptions.ServerResponseException;
+import com.contrastsecurity.exceptions.UnauthorizedException;
+import java.io.IOException;
+import java.nio.file.Path;
+
+/**
+ * Code artifact resource collection.
+ *
+ * <p>This resource API entry point does not use the {@code define()} pattern for creating new code
```
tylernhoward, line 13: I recognized this was really the only resource collection without this pattern and was going to ask.  👍

src/main/java/com/contrastsecurity/sdk/scan/ProjectClient.java
```diff
@@ -0,0 +1,45 @@
...
+ *     every request is tedious. Instead, it is expected that implementations will provide an
+ *     organization ID.
+ */
+interface ProjectClient {
+
+  /**
+   * Creates a new Scan project.
+   *
+   * @param create new project request
+   * @return the new {@link Project}
```
tylernhoward, line 24: If i was being pedantic I would say `ProjectInner` is not of type `Project` technically - though they are functionally equivalent.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
tylernhoward, line 8: I'm not crazy about this naming convention, but I think `Inner` is better than `Value` and both are certainly better than `Form`. I don't have any suggestions for the time being - and i like that this pattern gets us around some GSON hiccups and ugliness

tylernhoward (approved):

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
tylernhoward: Bryan's naming convention is one I do like more as far as illustrating what the object really is, I do see the point regarding suffix vs prefix naming though. While `Api` is more descriptive imo it is unfortunately not fit for a suffix in a concise way. I can also see us going down a naming rabbit hole that is similar to what we currently have. We may as well just follow the pattern we are attempting to emulate for the time being

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactImpl.java
```diff
@@ -0,0 +1,61 @@
+package com.contrastsecurity.sdk.scan;
+
+import java.time.Instant;
+import java.util.Objects;
+
+/** Implementation of the {@link CodeArtifact} resource. */
+final class CodeArtifactImpl implements CodeArtifact {
```
tylernhoward, line 7: I'm sure there are some auto-delegation libraries that could potentially remove some of this boiler plate, one that comes to mind is: https://github.com/ryandens/auto-delegate
I've heard good things about this dev too!

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
gilday: Do you also think it makes sense to introduce an exception to the rule here, or just use `define().file(...).create()` anyway even though it's a bit wordy and annoying?

src/main/java/com/contrastsecurity/sdk/scan/ProjectClient.java
gilday: This is a copy paste error, I'll fix it.

src/main/java/com/contrastsecurity/exceptions/HttpResponseException.java
gilday: I'm testing with the Maven plugin now, and I'm finding that I'd like even more debug information in the exception message. I'm going to add the request path and method.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
bryanateoan: I'm fine with following the established pattern

src/main/java/com/contrastsecurity/sdk/scan/ProjectClient.java
gilday: This mistake is repeated in all of the clients 🤦

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactImpl.java
gilday: I would love to use this generator, but it doesn't fit this use case. It generates code for delegating to an instance in the same type hierarchy e.g.

```
class MySet implements Set {
}
```

PR #99, opened 2021-08-20
Set source file encoding to prevent warnings

sethjackson:
Stops the warnings about source file encodings when compiling.


gilday:
Good catch - we just merged a fix to `main` for this from another branch.

sethjackson:
👍

PR #104, opened 2021-08-27
INT-805 Fix server event to string

bryanateoan:
Changes the toString for ServerEnvironment to UPPERCASE so that teamserver will accept it. This is also to match the behavior with RuleSeverity.

### DIFF
--- a/src/test/java/com/contrastsecurity/http/TraceFilterFormTest.java
+++ b/src/test/java/com/contrastsecurity/http/TraceFilterFormTest.java
@@ -1,0 +1,58 @@
+package com.contrastsecurity.http;
+/*-
+ * #%L
+ * Contrast Java SDK
+ * %%
+ * Copyright (C) 2021 Contrast Security, Inc.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/contrastsecurity/http/ServerEnvironment.java
```diff
@@ -20,21 +20,27 @@
...
   }
 
+  public String toURIString() {
+    return this.label.toUpperCase();
+  }
+
   @Override
   public String toString() {
-    return this.label.toLowerCase();
+    return toURIString();
```
gilday, line 44: I assume that GSON calls `toString()` when serializing enums, and that's why this fixes the issue. Is that right?

gilday (approved):

src/main/java/com/contrastsecurity/http/ServerEnvironment.java
bryanateoan: The TraceFilterForm and ApplicationFilterForm calls toString() when creating the uri parameters.

```Java
if (environments != null && !environments.isEmpty()) {
      filters.add(
          "environments="
              + environments.stream().map(Object::toString).collect(Collectors.joining(",")));
    }
```

gilday:
I see. We should update that to toURIString
________________________________

bryanateoan:
I updated the filterforms to use toURIString and I also added a toURIString for RuleSeverity

src/main/java/com/contrastsecurity/http/TraceFilterForm.java
```diff
@@ -123,7 +127,10 @@ public String toQuery() throws UnsupportedEncodingException {
 
     if (expand != null && !expand.isEmpty()) {
       filters.add(
-          "expand=" + expand.stream().map(Object::toString).collect(Collectors.joining(",")));
```
arturoespinoza15, line 126: do you think we need a test in this method `toQuery()` to check if the filter string is built as expected?

bryanateoan: Added tests.

arturoespinoza15 (approved):

bryanateoan:
reminder for review when you get a chance. getting a couple of pokes for this one.

tylernhoward (approved):

gilday:
😕 I thought I already approved. I'll do it again!

gilday (approved):

PR #105, opened 2021-08-27
update build java version to 16

bryanateoan:
Looks like temurin only has 8, 11, and 16. Java 16 is not compatible with the latest versions of spotless and javadoc plugins. spotless has a workaround but we can't use it since we are targeting java 8.

Setting the build workflow to use 11, until our dependencies get updated to 16.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v2
         with:
-          java-version: 15
+          java-version: 16
           distribution: temurin
       - name: Cache Maven Wrapper
         uses: actions/cache@v2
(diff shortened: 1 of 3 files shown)


gilday:
We can workaround the Spotless issue, but I have no solution for the incompatibility with the doclet we're using for code samples

```xml
<docletArtifact>
  <groupId>org.apidesign.javadoc</groupId>
  <artifactId>codesnippet-doclet</artifactId>
  <version>0.61</version>
</docletArtifact>
```

It's using internal classes that have been hidden in JDK 16.

There's a JEP for introducing code snippets to the javadoc tool natively (see https://bugs.openjdk.java.net/browse/JDK-8201533). It's slated for JDK 18. I recommend we use only JDK 11 in our build chain (continuing to target JDK 1.8 with the `--release` javac flag) until JDK 18 at which time we can use the native code snippet support.

PR #107, opened 2021-09-02
Support expanding servers for libraries

sethjackson:
This adds the ability to expand servers on libraries. Previously servers were expanded when expanding apps but now an additional expansion was added for servers.


sethjackson:
@bryanateoan could someone review this one? It's to support API changes in newer TeamServer versions.

sethjackson:
Thanks!

PR #109, opened 2021-11-04
add metadata parameter to CodeArtifact api

seschis:


src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
```diff
@@ -43,12 +43,14 @@
...
+   *     prescan data is not present.
    * @return new {@link CodeArtifactInner} from Contrast API
    * @throws IOException when an IO error occurs while making the request to the Contrast API
    * @throws UnauthorizedException when Contrast rejects the credentials used to send the request
    * @throws ResourceNotFoundException when the requested resource does not exist
    * @throws HttpResponseException when Contrast rejects this request with an error code
    * @throws ServerResponseException when Contrast API returns a response that cannot be understood
    */
-  CodeArtifactInner upload(String projectId, Path file) throws IOException;
+  CodeArtifactInner upload(String projectId, Path file, Path metadata) throws IOException;
```
seschis, line 55: should this function signature been added as an additional one rather than replacing the original one for any kinda of backwards compat?

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifact.java
```diff
@@ -40,6 +41,10 @@
   /** @return filename */
   String filename();
 
+  @Nullable
+  /** @return metadata filename */
```
bryanateoan, line 45: can you put the annotation bellow the comment?

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactInner.java
```diff
@@ -44,6 +45,10 @@ static Builder builder() {
   /** @return filename */
   abstract String filename();
 
+  @Nullable
+  /** @return metadata filename */
```
bryanateoan, line 49: can you put the annotation bellow the comment?

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
bryanateoan: is it valid use-case to upload a file without metadata?

bryanateoan (changes requested):

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
gilday, line 55: This API asks that the user provide the metadata JSON as a file. Maybe passing the metadata as a file is the only use case we have now (i.e. the Maven plugin), but it doesn't feel like the most flexible assumption for the SDK. I would expect the SDK to accept the metadata as a Java object that describes the data structure, and I would expect the SDK to marshal that object to JSON for me.

Are the fields in the metadata JSON object well known, or is this more of an open-ended bag of key-value pairs? If it's the former, then we should define a new class for it. If it's the latter, then maybe all we need is a `Map`.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifact.java
gilday: @bryanateoan consider adding Checkstyle to this project to catch things like this. See https://checkstyle.sourceforge.io/config_javadoc.html#InvalidJavadocPosition

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
gilday: According to the Javadoc, it is valid to upload a file without metadata. Therefore, I think we should add a method override that introduces this parameter.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClientImpl.java
```diff
@@ -80,8 +81,31 @@ public CodeArtifactInner upload(final String projectId, final Path file) throws
             + "Content-Transfer-Encoding: binary"
             + CRLF
             + CRLF;
-    final String footer = CRLF + "--" + boundary + "--" + CRLF;
-    final long contentLength = header.length() + Files.size(file) + footer.length();
+    final String metadataSection;
+    if (metadata != null) {
```
gilday, line 85: Consider a ternary expression.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactImpl.java
```diff
@@ -52,6 +53,12 @@ public String filename() {
     return inner.filename();
   }
 
+  @Override
+  @Nullable
```
gilday, line 57: Do you need `@Nullable` here? I believe it should be inherited from the interface definition.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
```diff
@@ -62,6 +62,7 @@
    */
   CodeArtifact upload(Path file, String name) throws IOException;
 
+  CodeArtifact upload(Path file, String name, Path metadata, String metaname) throws IOException;
```
gilday, line 65: What is `metaname` here? I don't think it's used.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
```diff
@@ -75,4 +76,6 @@
    * @throws ServerResponseException when Contrast API returns a response that cannot be understood
    */
   CodeArtifact upload(Path file) throws IOException;
+
+  CodeArtifact upload(Path file, Path metadata) throws IOException;
```
gilday, line 80: Needs a Javadoc comment.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
```diff
@@ -65,6 +68,10 @@ void before(@TempDir final Path tmp) throws IOException {
     try (JarOutputStream jos = new JarOutputStream(new FileOutputStream(jar.toFile()))) {
       jos.putNextEntry(new ZipEntry("HelloWorld.class"));
     }
+    metadataJson = tmp.resolve("prescan.json");
+    try (FileOutputStream fos = new FileOutputStream(metadataJson.toFile())) {
```
gilday, line 72: Typically, when you're writing character data, you want to wrap that `FileOutputStream` with an `OutputStreamWriter` and have the writer take care of the character encoding, but since this is just a small test, the `"literal".getBytes(StandardCharsets.UTF_8)` approach is probably fine.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
```diff
@@ -78,7 +85,7 @@ RequestResponsePact pact(final PactDslWithProvider builder) throws IOException {
       params.put("organizationId", "organization-id");
       return builder
           .given("Projects Exist", params)
-          .uponReceiving("upload new code artifact")
+          .uponReceiving("upload new code artifact with metadata")
```
gilday, line 88: there's no metadata in this test.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
```diff
@@ -115,14 +122,91 @@ void upload_code_artifact(final MockServer server) throws IOException {
...
+                            TestDataConstants.TIMESTAMP_EXAMPLE);
+                      })
+                  .build())
+          .toPact();
+    }
+
+    @Disabled("https://github.com/pact-foundation/pact-jvm/issues/668")
+    @Test
+    void upload_code_artifact_with_metadata(final MockServer server) throws IOException {
+      System.out.println("running metadata test");
```
gilday, line 194: 🔥 this println

gilday (changes requested):
Mostly there, but I'd like to figure out a better API for users to pass the metadata to the SDK.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
seschis: `metaname is analagous to the `name` argument the way `metadata` is analogous to the `file` argument.  As to if that is even needed, I'm not sure.  I don't see why we'd give the user the capability to change the "filename" on the multipart upload... but I figured it was something I didn't understand so I just mirrored the behavior when allowing the prescan metadata to be added to the multipart file upload.  It's your call if you want me to take it out as I don't have a valid argument for or against it.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
seschis: whoops, refactoring junk that I missed.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
seschis: lol, some habits....

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
seschis: the format of the data is defined here: https://github.com/Contrast-Security-OSS/contrast-scan-prescan/blob/master/src/main/resources/schema/scan-input-metadata-schema-1.0.0.json

I don't anticipate it changing from json, but I can't say it never will.  I also don't know how the content of the data will be required to change over time.  The main purpose of the prescan metadata is to allow the engine to generate physical absolute paths to files in its sarif report which GitHub uses as part of preview renderings when displaying the sarif findings.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactImpl.java
seschis: I removed this and everything still works... so I guess not.

src/test/java/com/contrastsecurity/sdk/scan/CodeArtifactsPactTest.java
seschis: thanks for the idiomatic pattern, still learning to be a Java dev :).  I changed to match the suggestion.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClient.java
gilday: I understand. Knowing that has a well defined schema, I feel more strongly that we should make a corresponding Java class to hold this data, then the SDK can turn that object (that the SDK user supplies to this API) into JSON. Does that make sense?

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClientImpl.java
```diff
@@ -80,8 +91,28 @@ public CodeArtifactInner upload(final String projectId, final Path file) throws
             + "Content-Transfer-Encoding: binary"
             + CRLF
             + CRLF;
-    final String footer = CRLF + "--" + boundary + "--" + CRLF;
-    final long contentLength = header.length() + Files.size(file) + footer.length();
+    final String metadataSection =
+        metadata != null
+            ? boundaryMarker
+                + CRLF
+                + "Content-Disposition: form-data; name=\"metadata\"; filename=\""
```
gilday, line 98: If we know this form is of type application/json, shouldn't we send it as such instead of sending it as a file? This relates to my question in `CodeArtifactClient` which is "how does the user know how to build this file?"

gilday (commented):
All of my remaining comments pertain to how we're sending the metadata. On one hand, the metadata is defined by a well-known JSON schema, so I feel that it should be represented by a Java class, and its representation on the wire as JSON should be abstracted from the SDK user. On the other hand, I'm seeing some evidence that the file is an opaque box to the user, and they should know or care that it's JSON.

I can't tell which is our intention. I think learning more about the API and the way that this file is generated would help me understand.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClientImpl.java
seschis: I think this gets back to the API design of the REST endpoint which I did not design.  Let me see if I can provide some clarity on a few of your questions for trying to figure the best way to proceed.

The JSON spec I linked to (https://github.com/Contrast-Security-OSS/contrast-scan-prescan/blob/master/src/main/resources/schema/scan-input-metadata-schema-1.0.0.json) is an ingest specification for the scanner engine itself.  It's saying, that if you want to provide the scanner engine some "prescan data" this is the format it will accept it in.  Its published this way because I just wanted a specification for the input format rather than telling someone, "go read the code", and I also wanted to use code generators like jsonschema2pojo, to automatically generate the libraries to handle the data models of the scanner's prescan ingest format. 

* The scanner makes no promises that this format will **always** be supported, though we will try to keep it backwards compat best as possible, in all future versions of the scanner engine.  The current format represents the simplest format to satisfy the current requirements given the current knowledge of the domain, which we know is not complete.
* The REST API in front of the scanner engine doesn't do anything with this data specifically and merely passes it along to the scanner engine.  From its point of view, it's opaque.
* If the REST api wanted to validate the data before passing to the engine since it knows what engine version it bundles, its free to do that with the publish json schema, but the engine makes no assumptions that will occur.
* For "contrast tools" created to generate prescan data as part of the scan request, they have a well defined json-schema to use to get the right data to the engine. But once again, the engine makes no assumptions that it is guarded behind a REST api or anything thing.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifacts.java
seschis: ok, I see now.  I can take it out.  Before I do any work on it yet though I'll wait until we resolve the fundamental API design questions you have in this area.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactImpl.java
```diff
@@ -52,6 +52,11 @@ public String filename() {
     return inner.filename();
   }
 
+  @Override
+  public String metadata() {
```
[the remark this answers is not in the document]
seschis: hmm, I always considered the generation of prescan data something that was external to the SDK.... but it probably would be more user-friendly to allow the SDK user to call a function that generated prescan data for them.... or perhaps just do it transparently.

src/main/java/com/contrastsecurity/sdk/scan/CodeArtifactClientImpl.java
gilday: > For "contrast tools" created to generate prescan data as part of the scan request

Is this SDK an example of such a "contrast tool", or does the SDK user have to use another tool to generate the file first before they can use the SDK to include in their new code artifacts?

PR #112, opened 2022-02-24
INT-1102 update copyright dates

bryanateoan:
# Overview
This updates the copyright dates and allows the plugin to automatically make updates to the copyright section. This will automatically update the copyright dates in a new year. 

# Changes
1. Updated inceptionYear to be the same as when version 1.0 was released
2. Updated file headers
3. Updated license file
4. Allowed license plugin to make updates to the copyright section of the header

# Notes
The canUpdateCopyright is defaulted to false because the developer should be updating the copyright dates. https://www.mojohaus.org/license-maven-plugin/update-file-header-mojo.html#canUpdateCopyright. I think it is fine for us to set it to be true because the headers are updated only in the developer profile and we use PRs, so it will be difficult to accidentally update when you don't want it to. The alternative is to delete the headers in each file either by the `remove-file-header` goal or manually, and then rerun to update the header as a whole. 

There is a `update-project-file` https://www.mojohaus.org/license-maven-plugin/update-project-license-mojo.html to automatically update the license file but it had a bunch of stuff that I did not understand so I didn't touch it and just updated the LICENSE file manually.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2014 - 2022 Contrast Security, Inc.
+Copyright 2022 Contrast Security, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2021 Contrast Security, Inc.
+Copyright 2014 - 2022 Contrast Security, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,7 @@
     <versions.junit-jupiter>5.7.1</versions.junit-jupiter>
     <versions.auto-value>1.8.2</versions.auto-value>
     <versions.lombok>1.18.18</versions.lombok>
+    <copyrightYear>2022</copyrightYear>
   </properties>
 
   <dependencies>
@@ -354,7 +355,7 @@
           <version>2.0.0</version>
           <configuration>
             <licenseName>apache_v2</licenseName>
-            <inceptionYear>2014</inceptionYear>
+            <inceptionYear>${copyrightYear}</inceptionYear>
             <canUpdateCopyright>true</canUpdateCopyright>
           </configuration>
         </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -354,7 +354,8 @@
           <version>2.0.0</version>
           <configuration>
             <licenseName>apache_v2</licenseName>
-            <inceptionYear>2021</inceptionYear>
+            <inceptionYear>2014</inceptionYear>
+            <canUpdateCopyright>true</canUpdateCopyright>
           </configuration>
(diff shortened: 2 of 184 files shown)


cgperez (approved):

MissaelHRC (approved):

arturoespinoza15 (approved):

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2021 Contrast Security, Inc.
+Copyright 2014 - 2022 Contrast Security, Inc.
```
gilday, line 1: I don't think we're supposed to have this range in here. I'll send you the internal guidance I'm working off of.

pom.xml
```diff
@@ -354,7 +354,8 @@
           <version>2.0.0</version>
           <configuration>
             <licenseName>apache_v2</licenseName>
-            <inceptionYear>2021</inceptionYear>
+            <inceptionYear>2014</inceptionYear>
```
gilday, line 357: I don't think we're supposed to have an inception year in the headers. I'll send you the internal guidance I'm working off of.

gilday (changes requested):

LICENSE
bryanateoan: addressed

pom.xml
bryanateoan: addressed

gilday (approved):
lgtm

While I was skimming through the changes, I noted that we have a bunch of files that have old copyrights in them as well. It seems that the old copyrights are above the `package` declaration, so we missed them the first time we went through this exercise.

Consider chopping out those old, conflicting copyrights now.

```bash
$ grep -Rl 'Copyright (c) 201'
./src/main/java/com/contrastsecurity/models/CodeObject.java
./src/main/java/com/contrastsecurity/models/Application.java
./src/main/java/com/contrastsecurity/models/Traces.java
./src/main/java/com/contrastsecurity/models/NameValuePair.java
./src/main/java/com/contrastsecurity/models/Coverage.java
./src/main/java/com/contrastsecurity/models/Library.java
./src/main/java/com/contrastsecurity/models/User.java
./src/main/java/com/contrastsecurity/models/Login.java
./src/main/java/com/contrastsecurity/models/TraceEvent.java
./src/main/java/com/contrastsecurity/models/TracesWithResponse.java
./src/main/java/com/contrastsecurity/models/Servers.java
./src/main/java/com/contrastsecurity/models/Users.java
./src/main/java/com/contrastsecurity/models/RouteCoverageResponse.java
./src/main/java/com/contrastsecurity/models/Server.java
./src/main/java/com/contrastsecurity/models/Trace.java
./src/main/java/com/contrastsecurity/models/HttpRequest.java
./src/main/java/com/contrastsecurity/models/URLEntry.java
./src/main/java/com/contrastsecurity/models/SignUp.java
./src/main/java/com/contrastsecurity/exceptions/ResourceNotFoundException.java
./src/main/java/com/contrastsecurity/exceptions/InvalidConversionException.java
./src/main/java/com/contrastsecurity/exceptions/UnauthorizedException.java
./src/main/java/com/contrastsecurity/sdk/ContrastSDK.java
```

PR #116, opened 2022-09-02
TS-18990 Fix Tracked and Untracked Filters on Contrast SDK Java Library

MiguelUrrutiaSoler:
Fixed Status adding new flags for tracked and untracked

### DIFF
--- a/src/main/java/com/contrastsecurity/http/TraceFilterForm.java
+++ b/src/main/java/com/contrastsecurity/http/TraceFilterForm.java
@@ -158,10 +158,7 @@ public class TraceFilterForm {
     }
 
     if (status != null) {
-      List<String> filteredStatus = status.stream()
-              .filter(x -> x != "tracked" && x != "tracked")
-              .collect(Collectors.toList());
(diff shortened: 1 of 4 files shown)


src/main/java/com/contrastsecurity/http/TraceFilterForm.java
```diff
@@ -92,6 +92,8 @@ public String toString() {
   private int limit;
   private int offset;
   private String sort;
+  private Boolean tracked;
```
dskelly-contrast, line 95: These seem like they could just be primitive bools?

src/main/java/com/contrastsecurity/http/TraceFilterForm.java
```diff
@@ -92,6 +92,8 @@ public String toString() {
...
+  private Boolean untracked;
```
dskelly-contrast, line 96: Can we add some comments indicating what tracked and untracked stand for? Just so future code readers are not confused.

dskelly-contrast (approved):

garellano (approved):

PR #117, opened 2024-10-11
Java 3733 scaffold for gradle plugin

BrianPhillips2020:
Scaffolding for the new gradle plugin. Registers simple tasks, one hello world and one no-op (partly for my person edification on how task registration works). There are a few things to finish/clear up before this is merged

- [x] Project structure: The project is currently structured like this
```
contrast-sdk-java
|
├── gradle-plugin
|     ├──   build.gradle
|     ├──  settings.gradle
|     └── src
|           └── ...
|
├── src
|   └── ...
├── mvnw
└── pom.xml
```
Should it instead be structured so the `gradle-plugin` and the mvn build for the sdk are in completely separate projects? such as:

```
contrast-sdk-java
|
├── gradle-plugin
|     ├──   build.gradle
|     ├──  settings.gradle
|     └── src
|           └── ...
├── sdk
      ├── src
      |  └── ...
      ├── mvnw
      └── pom.xml
```

- [x] Need to implement branch protections so code cannot be pushed directly to main. Enforced through github? I need to check this out

- [x] Decide how the build pipeline should work. I think separate both the mvn build and the gradle build into different actions have a single workflow relying on both?


gradle-plugin/build.gradle
stiemannkj1, line 1: Are people still building on Java 8? I think we should set the compiler release level to 8. Or at least 11.

.github/workflows/build-gradle-plugin.yml
```diff
@@ -0,0 +1,29 @@
...
+      - main
+  pull_request:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
stiemannkj1, line 16: You need to use the full has here to be safe from supply chain attacks.

.github/workflows/build-gradle-plugin.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: echo "The ${{ github.repository }} repo has been cloned to the runner
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
```
stiemannkj1, line 20: You need to use the full has here to be safe from supply chain attacks.

.github/workflows/build-gradle-plugin.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Setup Java
+        uses: actions/setup-java@v2
```
stiemannkj1, line 23: You need to use the full has here to be safe from supply chain attacks.

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/RegistrationTests.java
stiemannkj1, line 1: Is there really any value in these tests if we have integration tests? I guess keep them for now, but I think we might end up deleting them. I'd rather validate with the tests that actually run stuff.

stiemannkj1 (approved):

stiemannkj1:
Nice work!

stiemannkj1:
> Should it instead be structured so the gradle-plugin and the mvn build for the sdk are in completely separate projects? 

Yes, please make that change too.

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/RegistrationTests.java
BrianPhillips2020: We'll probably end up deleting them. This is the most basic smoke test you can make for a gradle plugin. Everything I've seen in researching this has this little test just verifying they're registered. Might be worth having just for the very very simple case of identifying if registration breaks, but I'm not attached

gradle-plugin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+group = 'com.contrastsecurity.gradle.plugin'
+
+repositories {
+    mavenCentral()
+}
+
+gradlePlugin {
+    plugins {
+        java {
+            id = 'com.contrastsecurity.java'
```
rszewczyk, line 17: How are we choosing the plugin name?

gradle-plugin/build.gradle
rszewczyk: @stiemannkj1 are you saying we _should_ support building as far back as Java 8 or 11? Why though?

rszewczyk (approved):

rszewczyk (approved):

gradle-plugin/build.gradle
rszewczyk: 🤔 It looks like Gradle [still supports Java 8](https://docs.gradle.org/current/userguide/compatibility.html), so long as you're not using the daemon, even on it's latest versions, so I guess that is prudent.

gradle-plugin/build.gradle
BrianPhillips2020: I forgot to bring this up, no idea 🙃 nobody picked a name yet, this is completely arbitrary. Easily changeable at least, I might make permanently choosing that a requirement for before we actually release

rszewczyk: Yeah, it's not something we need to decide until we publish. Calling the plugin "java" seems a bit vague, but I can't think of something better at the moment. 🤷

stiemannkj1:
Added branch protection for `main` myself since I don't think you have permissions.

PR #121, opened 2025-01-13
Java 8886 agent input and skd provider

BrianPhillips2020:
# Purpose
Redesigned the plugin to be more inline with Gradle best practices

# Tasks
## ResolveAgent
- Input: Provided path to agent
- Output: The agent file
- DependsOn: None
- Description: Resolves the agent path if provided. If no path is provided or the file does not exist, uses the SDK to pull the agent down from TeamServer

## ContrastCheck
- Input: none
- Output: none
- DependsOn: `ResolveAgentTask`, all `contrastVerifyTest<corresponding-task-names>`
- Description: Lifecycle task which acts as a wrapper around all the `contrastVerifyTest` tasks. This is what the user will call when they want contrast analysis on their tests

## ContrastVerifyTest
- Input: It's corresponding Test tasks output, i.e. test results
- Output: TS data
- DependsOn: Corresponding `Test` task
- Description: Essentially anonymous Tasks. These Tasks have a one to one relationship with each registered `Test` task. Queries the TS results for its test and verifies that no new vulnerabilities have been found. TODO: implementation of the verification logic: https://contrast.atlassian.net/browse/JAVA-3779

# SDK Service
Initializes and provides a static instance for the Contrast SDK to be used throughout the plugin

# Gradle Plugin
Registers and configures tasks. Loops through the registered `Test` tasks for the project and generates our `ContrastVerifyTest` for each one.


BrianPhillips2020:
Still needs testing

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,140 @@ public void apply(final Project target) {
...
         target.getExtensions().create(EXTENSION_NAME, ContrastConfigurationExtension.class);
 
-    target.getTasks().register("installAgent", InstallAgentTask.class);
+    TaskProvider<ResolveAgentTask> resolveAgentTask =
+        target.getTasks().register("resolveAgent", ResolveAgentTask.class);
+
+    resolveAgentTask.configure(
+        task -> {
+          String jarPath;
+          if (extension.getJarPath() != null) {
```
argonotix, line 33: I would refactor this slightly for readability:

```java
String jarPath = extension.getJarPath();
jarPath = jarPath != null ? jarPath : toJarPath(target);

...

private static String toJarPath(Target target) {
 return target.getLayout().getBuildDirectory().getAsFile().get().getPath()
                    + "/contrast.jar";
}
```

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,140 @@ public void apply(final Project target) {
...
+                task -> {
+                  /*
+                  We need the agent before we can set the jvm args for the tests, and the jvm args must be set during configuration, so we must set this here rather than as on the test tasks.
+                   */
+                  task.dependsOn(resolveAgentTask);
+                  // temporary for debugging
+                  target
+                      .getTasks()
+                      .withType(Test.class)
+                      .forEach(t -> System.out.println(t.getAllJvmArgs()));
```
argonotix, line 59: Use gradle logger for debugging

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,140 @@ public void apply(final Project target) {
...
+    final String circleBuildNum = System.getenv("CIRCLE_BUILD_NUM");
+
+    final String appVersionQualifier;
+    if (travisBuildNumber != null) {
+      appVersionQualifier = travisBuildNumber;
+    } else if (circleBuildNum != null) {
+      appVersionQualifier = circleBuildNum;
+    } else {
+      appVersionQualifier = new SimpleDateFormat("yyyyMMddHHmmss").format(currentDate);
+    }
```
argonotix, line 159: nit ~ add whitespace

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,106 @@
...
+    this.fileString = fileString;
+  }
+
+  public void setAgent(File agent) {
+    this.agent = agent;
+  }
+
+  @TaskAction
+  void resolveAgent() {
+    System.out.println("running RESOLVE AGENT");
```
argonotix, line 52: gradle logger

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,106 @@
...
+    final ContrastSDK sdk = ContrastSDKService.getInstance(config);
+    agent = retrieveAgent(sdk, getFileString(), config.getOrgUuid(), getProject());
+  }
+
+  /** Use ContrastSDK to download agent and return the path where agent jar is stored */
+  @VisibleForTesting
+  public static File retrieveAgent(
```
argonotix, line 59: `throws GradleException`

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @TaskAction
+  void resolveAgent() {
+    System.out.println("running RESOLVE AGENT");
+    final ContrastSDK sdk = ContrastSDKService.getInstance(config);
+    agent = retrieveAgent(sdk, getFileString(), config.getOrgUuid(), getProject());
+  }
+
+  /** Use ContrastSDK to download agent and return the path where agent jar is stored */
```
argonotix, line 57: Update jdoc to reflect params and return value

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/extensions/ContrastSDKService.java
```diff
@@ -0,0 +1,39 @@
+package com.contrastsecurity.gradle.plugin.extensions;
+
+import com.contrastsecurity.sdk.ContrastSDK;
+import com.contrastsecurity.sdk.UserAgentProduct;
+
+public class ContrastSDKService {
```
argonotix, line 6: jdoc

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/extensions/ContrastSDKService.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public static ContrastSDK INSTANCE;
+
+  public static void initializeSdk(final ContrastConfigurationExtension extension) {
```
argonotix, line 10: jdoc

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,106 @@
...
+      } catch (FileAlreadyExistsException e) {
+        logger.debug("Project dir already exists");
+      }
+
+      downloadedAgent = target.resolve(jarPath);
+
+      Files.write(downloadedAgent, bytes, StandardOpenOption.CREATE, StandardOpenOption.WRITE);
+
+    } catch (RuntimeException | IOException e) {
+      throw new GradleException("Failed to download java agent from the Contrast api: " + e);
```
argonotix, line 98: I wonder if we should really throw an exception on this one...  I'm not sure we want to block a customer from running tests just because we weren't able to obtain an agent.

I think we should have some kind of fallback that emits a large message in the Gradle logs that the agent isn't being run or something.  I don't know how others feel about this but as written I'm pretty sure this will dump and block all testing.

- [ ] Verify that I'm right about that

argonotix (changes requested):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
BrianPhillips2020: The task needed to run this is separate from Test Tasks right now. Running `contrastCheck` runs all the tests and explicitly expects the agent to be attached. I think I've been swapping back and forth between auto connecting contrast with the plugin applied vs having a task to do it. This way makes more sense to me right now but I am open to discussing it

argonotix: If it's being run separately then this is fine.  It's only an issue if this was an 'always run' situation. I'd make sure this doesn't make things croak however like just refreshing Gradle for example.

BrianPhillips2020: You made me go back and check and actually you were right. The plugin configures tests to have the agent args always, because this happens in the plugin itself rather than a task. I'm conflicted again on how this should work. I think I'm going to experiment briefly with configuring the `Test` tasks within the `contrastCheck` task. I think this is antithetical to what Gradle wants with Lifecycle tasks but 🤷

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,129 @@ public void apply(final Project target) {
...
+              // generate ContrastVerifyTestTask for each test
+              TaskProvider<ContrastVerifyTestTask> verifyTask =
+                  target
+                      .getTasks()
+                      .register(
+                          "contrastVerifyTest" + testTask.getName(), ContrastVerifyTestTask.class);
+              verifyTask.configure(
+                  v -> {
+                    v.dependsOn(testTask);
+                    // set file input for
```
BrianPhillips2020, line 70: TODO in next ticket

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
BrianPhillips2020: https://contrast.atlassian.net/browse/JAVA-8962

BrianPhillips2020:
The pipeline keeps failing on the maven verify step. I don't know why yet

BrianPhillips2020:
> The pipeline keeps failing on the maven verify step. I don't know why yet

https://github.com/Contrast-Security-OSS/contrast-sdk-java/pull/122

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/extensions/ContrastSDKService.java
```diff
@@ -0,0 +1,44 @@
...
+public class ContrastSDKService {
+
+  public static ContrastSDK INSTANCE;
+
+  /**
+   * Initializes the SDK for the plugin. If configuration from the {@link
+   * ContrastConfigurationExtension} is missing or incorrect, provides a NOOP Instance
+   */
+  public static void initializeSdk(final ContrastConfigurationExtension extension) {
+    if (extension.getUsername() == null
```
aberman, line 16: Once you change these props to `Property` you'll see you have new methods which will allow you avoid testing on null (basically the methods from `Optional`, e.g. `isPresent()`)

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/GradlePluginTests.java
```diff
@@ -0,0 +1,48 @@
+package com.contrastsecurity.gradle.plugin;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.gradle.api.Project;
+import org.gradle.testfixtures.ProjectBuilder;
+import org.junit.jupiter.api.Test;
+
+/** Smoke tests for proper plugin registration */
+class GradlePluginTests {
```
aberman, line 10: Gradle plugin tests are far more complicated than this.  You're not really testing much here tbh.  You may want to write up a separate ticket to do it.  You have to actually load the plugin, execute the tasks, etc. to test it properly.

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/e2e/EndToEndTests.java
```diff
@@ -23,35 +15,16 @@
 public class EndToEndTests extends GradleRunnerTest {
```
aberman, line 15: Seems like this is just to verify getting the agent from team server works.  I would just do a unit test that calls whatever method you decide on in the plugin which goes and gets the agent.  Save an actual plugin test for the plugin test class.

aberman (changes requested):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,136 @@ public void apply(final Project target) {
     ContrastConfigurationExtension extension =
         target.getExtensions().create(EXTENSION_NAME, ContrastConfigurationExtension.class);
 
-    target.getTasks().register("installAgent", InstallAgentTask.class);
+    TaskProvider<ResolveAgentTask> resolveAgentTask =
+        target.getTasks().register("resolveAgent", ResolveAgentTask.class);
+
+    resolveAgentTask.configure(
+        task -> {
+          String jarPath = extension.getJarPath();
```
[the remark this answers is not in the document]
BrianPhillips2020: This is working, though I'm not sure how consistently. I have manual tests showing this at least. I realized last night I want to ensure this **all** happens after any `Test` tasks could be registered, so I may need to figure out some better enforcement for registration/configuration ordering.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/extensions/ContrastConfigurationExtension.java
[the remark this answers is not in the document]
BrianPhillips2020: I'll try rewriting it like this, the problem I was facing before was gradle complaining without the setters and getters in this way (from what I can tell, groovy would kinda do this under the hood), I never tried with declaring them as `Property`s explicitly, I'll give it a go

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,136 @@ public void apply(final Project target) {
...
+    target
+        .getTasks()
+        .withType(Test.class)
+        .forEach(
+            testTask -> {
+              testTask.dependsOn(resolveAgentTask);
+
+              // attach agent args
+              testTask.jvmArgs(
+                  createContrastArgs(
```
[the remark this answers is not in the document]
BrianPhillips2020: Ah, I had it out of the loop but because of something with the ordering of how it got the extension it wasn't populating correctly. This must be why

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,136 @@ public void apply(final Project target) {
...
+
+      appVersionQualifier = new SimpleDateFormat("yyyyMMddHHmmss").format(currentDate);
+    }
+
+    return appName + "-" + appVersionQualifier;
+  }
+
+  /** Generates the expected path to the Java Agent */
+  private String toJarPath(final Project target) {
+    return target.getLayout().getBuildDirectory().getAsFile().get().getPath() + "/contrast.jar";
```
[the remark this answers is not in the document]
BrianPhillips2020: neat ok

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/GradlePluginTests.java
BrianPhillips2020: https://contrast.atlassian.net/browse/JAVA-8969 <-  new ticket for fixing the tests. Feels slightly too big to put in to this pr as well. manual tests are at least verifying nothing crashes and burns

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/e2e/EndToEndTests.java
BrianPhillips2020: See https://contrast.atlassian.net/browse/JAVA-8969, we'll lump all the testing together

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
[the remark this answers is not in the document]
BrianPhillips2020: you were correct! though it only manifested once I updated the way the extension populated to use `Property`s

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,136 @@ public void apply(final Project target) {
...
+      final String serverName,
+      String appVersion) {
+
+    // List to preserve ordering of arguments
+    final Collection<String> args = new LinkedList<>();
+
+    args.add("-javaagent:" + agentPath);
+
+    // Use gradle project name in the even the appName is not set
+    if ("null".equals(appName) || appName == null) {
```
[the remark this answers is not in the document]
BrianPhillips2020: I rechecked, I don't think this is part of the spec. I think this was left over jank from some testing from the maven plugin, 🔥 as this logic is now basically handled in the extension itself

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -14,7 +24,136 @@ public void apply(final Project target) {
...
+  @VisibleForTesting
+  public static String computeAppVersion(final String appName) {
+    final Date currentDate = new Date();
+    final String travisBuildNumber = System.getenv("TRAVIS_BUILD_NUMBER");
+    final String circleBuildNum = System.getenv("CIRCLE_BUILD_NUM");
+
+    final String appVersionQualifier;
+
+    if (travisBuildNumber != null) {
+
```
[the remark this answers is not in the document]
BrianPhillips2020: Fixed the spacing a little, updated to still have extra whitespace but more readable hopefully

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class ContrastVerifyTestTask extends DefaultTask {
+
+  @InputFile @Optional File testResults;
+
+  public void setTestResults(final File testResults) {
+    this.testResults = testResults;
+  }
+
+  public File getTestResults() {
```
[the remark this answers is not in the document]
BrianPhillips2020: I ended up nuking all of these. I'm not going to worry about this yet because I'm going to write the actual implementation for this task in https://contrast.atlassian.net/browse/JAVA-3779. No need to figure out the test outputs here

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,97 @@
...
+ * <p>Input: Provided Agent Path
+ *
+ * <p>Output: The Java Agent jar file
+ */
+public class ResolveAgentTask extends DefaultTask {
+
+  final ContrastConfigurationExtension config =
+      (ContrastConfigurationExtension) getProject().getExtensions().getByName(EXTENSION_NAME);
+
+  @Input private String fileString;
```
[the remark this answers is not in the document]
BrianPhillips2020: why should the getters be abstract?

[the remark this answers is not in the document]
aberman: Gradle will tackle all the dirty work for you by creating an implementation.  You don't need any fields or anything, just the getters.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -1,21 +1,199 @@
...
+import org.gradle.api.tasks.testing.Test;
+import org.jetbrains.annotations.VisibleForTesting;
 
 /**
  * Gradle plugin for contrast utilities. The goals for this plugin are defined here <a
  * href=https://contrast.atlassian.net/browse/JAVA-8252>JAVA-8252</a>
  */
 public class ContrastGradlePlugin implements Plugin<Project> {
 
+  public static ContrastSDK SDK_INSTANCE;
```
aberman, line 26: This should not be static and I don't even think you need a field.  There's nothing stopping a customer from using the plugin in multiple projects where one project might have different extension values.  Each should be completely unique and pinned to the plugin instance of the specific project.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -1,21 +1,199 @@
...
+                        // set output file
+                        task.getAgent().set(new File(jarPath));
+                      });
+
+          // lifecycle task for grouping verification tasks
+          TaskProvider<Task> contrastCheck = target.getTasks().register("contrastCheck");
+
+          Collection<String> contrastArgs =
+              createContrastArgs(
+                  resolveAgentTask.get().getOutputs().getFiles().getSingleFile().getPath(),
```
aberman, line 58: When you do this, you are actually calling and executing the task and negating the value of it as a task (you should be able to see this when you debug that you'll actually have a value for this at this point in time when the plugin is applied).  I think you might be better off just nuking it as a task and just moving that code into this plugin.  What that means is that when the plugin loads and is applied, the code to download/determine the agent will also run.  The benefit there is it will only run once per plugin instance.  The downside is that it runs when the plugin loads, slowing down processing the build, but we have to do it somewhere.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/extensions/ContrastConfigurationExtension.java
```diff
@@ -0,0 +1,81 @@
...
+  private final Property<Boolean> attachToTests;
+
+  public ContrastConfigurationExtension(final Project project) {
+    username = project.getObjects().property(String.class);
+    apiKey = project.getObjects().property(String.class);
+    serviceKey = project.getObjects().property(String.class);
+    apiUrl =
+        project.getObjects().property(String.class).convention(ContrastGradlePlugin.DEFAULT_URL);
+    orgUuid = project.getObjects().property(String.class);
+    appName = project.getObjects().property(String.class).convention(project.getName());
```
aberman, line 27: Nice!

aberman (changes requested):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
BrianPhillips2020: Ok I understand this. I suppose the use case of a task output is to be used either after the plugin runs or as the input of another task. This isn't right because I want to use this task output in the application of the plugin. 

Perhaps we can figure out optimizations later for this. I can move this all into the plugin itself.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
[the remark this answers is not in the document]
BrianPhillips2020: I ended up removing this task and moving this all to the gradle plugin so this is no longer relevant anyway, but neat

gradle-plugin/src/test/java/com/contrastsecurity/gradle/plugin/e2e/EndToEndTests.java
```diff
@@ -1,114 +1,43 @@
...
 
 /** End-To-End tests for the gradle plugin for interacting with TeamServer */
-public class EndToEndTests extends GradleRunnerTest {
+public class EndToEndTests {
+
+  @TempDir public File tempDir = new File("/tempDir");
 
   @Test
-  void verify_retrieval_of_agent_from_teamserver() {
+  void it_verifies_agent_retrieval_from_TeamServer() throws IOException {
```
BrianPhillips2020, line 24: This test was moved here because it requires TS creds to work. Might think about renaming this package in the future but I think this unit test counts

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
BrianPhillips2020: This change was reverted. I put back the `ResolveAgentTask` and just updated it to be better because gradle would not allow us to write to the agent jar to the build directory without it. It's not worth our time right now to figure out why.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -11,11 +25,156 @@ public class ContrastGradlePlugin implements Plugin<Project> {
 
   public void apply(final Project target) {
 
-    ContrastConfigurationExtension extension =
+    final String agentPath =
```
argonotix, line 28: add comment that we're only using this as a target path. maybe that's kinda overkill but 🤷

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -11,11 +25,156 @@ public class ContrastGradlePlugin implements Plugin<Project> {
...
+   * @param appVersion the app version, if not supplied by the ContrastConfigurationExtension, then
+   *     it is generated in {@link ContrastGradlePlugin#computeAppVersion(String)}
+   * @return Set of arguments to be attached to specified tasks
+   */
+  @VisibleForTesting
+  public static Collection<String> createContrastArgs(
+      final String agentPath, final String appName, final String serverName, String appVersion) {
+
+    // List to preserve ordering of arguments
+    final Collection<String> args = new ArrayList<>();
```
argonotix, line 103: We can probably group these with `List.of` instead of adding one by one like this.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -11,11 +25,156 @@ public class ContrastGradlePlugin implements Plugin<Project> {
...
+
+    args.add("-javaagent:" + agentPath);
+
+    args.add("-Dcontrast.override.appname=" + appName);
+
+    args.add("-Dcontrast.server=" + serverName);
+
+    if (appVersion == null) {
```
argonotix, line 111: inline into a ternary

argonotix (approved):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
BrianPhillips2020: This actually isn't even a necessary check anymore thanks to the changes to the extension

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,96 @@
...
+    // Check if an agent was provided, and if so write that to the expected location
+    if (getProvidedPath().isPresent()) {
+      final String providedAgent = getProvidedPath().get();
+      Path providedPath = Paths.get(providedAgent);
+      if (!Files.exists(providedPath)) {
+        logger.debug("Unable to find java agentPath at " + providedAgent);
+
+      } else {
+        logger.debug("Agent provided via configuration retrieved");
+        writeToAgent(Files.readAllBytes(providedPath), agentBuildPath);
```
aberman, line 68: Do you want to return after this?  Right now it's going and still getting it from Team Server and calling writeToAgent again.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
```diff
@@ -0,0 +1,96 @@
...
+
+    // Check if an agent was provided, and if so write that to the expected location
+    if (getProvidedPath().isPresent()) {
+      final String providedAgent = getProvidedPath().get();
+      Path providedPath = Paths.get(providedAgent);
+      if (!Files.exists(providedPath)) {
+        logger.debug("Unable to find java agentPath at " + providedAgent);
+
+      } else {
+        logger.debug("Agent provided via configuration retrieved");
```
aberman, line 67: Little confused by this.  So, if the provided path does have a file, why do we write it again?  Can't we just use that file?  Seems like wasted i/o.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -1,7 +1,21 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.List;
+import java.util.Locale;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.gradle.api.tasks.TaskProvider;
+import org.gradle.api.tasks.testing.Test;
+import org.jetbrains.annotations.VisibleForTesting;
```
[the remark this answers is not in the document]
BrianPhillips2020: whoops force of habit

gradle-plugin/build.gradle
```diff
@@ -22,7 +22,7 @@ gradlePlugin {
 }
 
 compileJava {
-    options.release.set(8)
+    options.release.set(9)
```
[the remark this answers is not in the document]
BrianPhillips2020: I tried using `List.of()` and gradle wouldn't let me compile. I could put it back but idk if it matters too much? idk either way

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ResolveAgentTask.java
BrianPhillips2020: > Little confused by this. So, if the provided path does have a file, why do we write it again? Can't we just use that file?

This was like this because of a series of annoying restrictions involving the way gradle outputs work and the problems I had writing to the buildDir. You are right though, I don't think this writing to the other file is necessary. This new solution is a little hackier, but I think it's better, it checks for the provided agent twice, once in the plugin to set the task output, and once in the `ResolveAgentTask` to determine if it needs to get one from TS. It's annoying duplicate work though.

gradle-plugin/build.gradle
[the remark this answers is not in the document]
aberman: Ah yes, `List.of` is for Java 9+.   I would make sure this all compiles and works in Java 8 as it's still the most used Java AFAIK.  Remove the use of `List.of` and avoid any Java 9+ APIs

aberman (approved):
Approved pending comment

PR #122, opened 2025-01-16
JAVA-8967 update file headers for Lombok

BrianPhillips2020:
I spent an hour trying to figure out why lombok was failing mvn verify steps on my other pr, turns out it needed to _update the headers to say 2025_

Happy New Year Brian!


tylernhoward (approved):
good catch, we should figure out how to automate this as we have for our private repos.

PR #123, opened 2025-01-28
Java 3779 fail builds based on ts results

BrianPhillips2020:
Pulls Test results from TeamServer and fails the build if any traces are found which exceed or meet the min severity.

### DIFF
--- a/gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
+++ b/gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
@@ -1,15 +1,212 @@
 package com.contrastsecurity.gradle.plugin;
 
+import com.contrastsecurity.exceptions.UnauthorizedException;
+import com.contrastsecurity.gradle.plugin.extensions.ContrastConfigurationExtension;
+import com.contrastsecurity.http.RuleSeverity;
+import com.contrastsecurity.http.ServerFilterForm;
+import com.contrastsecurity.models.Application;
+import com.contrastsecurity.models.Applications;
+import com.contrastsecurity.models.Servers;
+import com.contrastsecurity.models.Trace;
+import com.contrastsecurity.models.TraceFilterBody;
+import com.contrastsecurity.models.Traces;
+import com.contrastsecurity.sdk.ContrastSDK;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Objects;
+import javax.inject.Inject;
 import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.logging.Logger;
 import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.TaskAction;
 
 /**
  * Task for verifying test results for its test. There is one of these tasks created for each Test
- * task in the project. This task is meant to take in it's corresponding Test task's output, use the
- * SDK to parse the TS results for the given task, then output a file corresponding to the results
- * found from TS.
+ * task in the project.
  *
- * <p>TODO: <a href="https://contrast.atlassian.net/browse/JAVA-3779">Implementation Ticket</a>
+ * <p>This task retrieves TS results for it's corresponding Test task, and if any vulnerabilities
+ * are found which match or exceed the configured minimum severity level, this task fails the build.
  */
 @CacheableTask
-public abstract class ContrastVerifyTestTask extends DefaultTask {}
+public abstract class ContrastVerifyTestTask extends DefaultTask {
+
+  final ContrastSDK sdk;
+
+  @Inject
+  public ContrastVerifyTestTask(ContrastSDK sdk) {
+    super();
+    this.sdk = sdk;
+  }
+
+  /** The trace results retrieved from TeamServer */
+  @OutputFile
+  public abstract RegularFileProperty getTraceResults();
+
+  @TaskAction
+  public void verifyFindings() {
+    final Logger logger = getProject().getLogger();
+
+    final ContrastConfigurationExtension config =
+        getProject().getExtensions().getByType(ContrastConfigurationExtension.class);
+
+    if (!config.getOrgUuid().isPresent()) {
+      throw new GradleException("Organization UUID not set");
+    }
+
+    final String orgUuid = config.getOrgUuid().get();
+    final String appId = getApplicationId(config, orgUuid);
+    final String serverName = config.getServerName().get();
+
+    final List<String> serverIds = getServerId(sdk, appId, orgUuid, serverName);
+
+    final TraceFilterBody traceFilter =
+        getTraceFilterBody(serverIds, config.getMinSeverity().get(), config.getAppVersion().get());
+
+    Traces traces;
+    try {
+      traces = sdk.getTraces(orgUuid, appId, traceFilter);
+    } catch (IOException e) {
+      throw new GradleException("Failed to retrieve traces from TeamServer", e);
+    }
+
+    if (traces != null && traces.getCount() > 0) {
+      logger.info(traces.getCount() + " new vulnerability(s) were found.");
+
+      // Write TeamServer results to a file in the build directory
+      try (BufferedWriter writer =
+          new BufferedWriter(new FileWriter(getTraceResults().get().getAsFile(), true))) {
+        for (Trace trace : traces.getTraces()) {
+          final String traceReport = generateTraceReport(trace);
+          writer.write(traceReport);
+          writer.newLine();
+          logger.lifecycle(traceReport);
+        }
+      } catch (final IOException e) {
+        logger.error("Could not write Contrast Findings to build directory", e);
+      }
+
+      throw new GradleException(
+          "Your application is vulnerable. Please see the above report for new vulnerabilities.");
+    } else {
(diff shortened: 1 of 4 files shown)


BrianPhillips2020:
I'm settling for manual testing right now because of the Q4 deadline and [this ticket](https://contrast.atlassian.net/browse/JAVA-8969), but can add at least unit tests if desired.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      logger.info(traces.getCount() + " new vulnerability(s) were found.");
+
+      // Write TeamServer results to a file in the build directory
+      try (BufferedWriter writer =
+          new BufferedWriter(new FileWriter(getTraceResults().get().getAsFile(), true))) {
+        for (Trace trace : traces.getTraces()) {
+          final String traceReport = generateTraceReport(trace);
+          writer.write(traceReport);
+          writer.newLine();
+          System.out.println(traceReport);
```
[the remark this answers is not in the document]
stiemannkj1: Probably want to use the gradle logger to log this at error I think.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      final ContrastConfigurationExtension config, final String orgUuid) {
+    Applications applications;
+
+    if (!config.getAppName().isPresent()) {
+      throw new GradleException("AppName or Organization UUID not set");
+    }
+    try {
+      applications = sdk.getApplications(orgUuid);
+    } catch (IOException e) {
+      throw new GradleException("Unable to retrieve the applications: " + e);
```
stiemannkj1, line 160: Don't do `+ e` for these exceptions. We want to provide `e` as the cause so it'll show up when users do `--stacktrace`.

```
throw new GradleException("Unable to retrieve the applications: ", e);
```

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      throw new GradleException("Unable to connect to TeamServer: " + e);
+    }
+
+    for (Application application : applications.getApplications()) {
+      if (Objects.equals(config.getAppName().get(), application.getName())) {
+        return application.getId();
+      }
+    }
+
+    throw new GradleException("Application with name '${applicationName}' not found.");
```
stiemannkj1, line 171: Need to provide the app name in this string.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+   * @param trace Trace object
+   * @return String report
+   */
+  private String generateTraceReport(final Trace trace) {
+    return "Trace: "
+        + trace
+            .getTitle()
+            // TODO check regex
+            .replaceAll("\\{\\{\\#unlicensed\\}\\}", "(")
+            .replaceAll("\\{\\{\\/unlicensed\\}\\}", ")")
```
stiemannkj1, line 197: No need for a regex here. If you're replacing the literal strings `{{#unlicensed}}` and {{/unlicensed}}`, use `String.replace`.

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
BrianPhillips2020: right, thanks this is not groovy

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+    List<String> severityList = SEVERITIES.subList(SEVERITIES.indexOf(severity), SEVERITIES.size());
+
+    List<RuleSeverity> ruleSeverities = new ArrayList<RuleSeverity>();
+
+    for (String severityToAdd : severityList) {
+      ruleSeverities.add(RuleSeverity.valueOf(severityToAdd.toUpperCase()));
+    }
+
+    return ruleSeverities;
+  }
```
stiemannkj1, line 224: I think there's probably a simpler way to do this with the `RuleSeverity` enums. Probably can just use the array of values and iterate over that starting at the lowest severity.

stiemannkj1 (approved):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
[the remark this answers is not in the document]
aberman: Do you always want the report to be spit out?  If so, use `logger.lifecycle`

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      try (BufferedWriter writer =
+          new BufferedWriter(new FileWriter(getTraceResults().get().getAsFile(), true))) {
+        for (Trace trace : traces.getTraces()) {
+          final String traceReport = generateTraceReport(trace);
+          writer.write(traceReport);
+          writer.newLine();
+          System.out.println(traceReport);
+        }
+      } catch (final IOException e) {
+        logger.debug("Could not write Contrast Findings to build directory: " + e);
```
aberman, line 95: Shouldn't this be `logger.error`?

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      }
+
+      throw new GradleException(
+          "Your application is vulnerable. Please see the above report for new vulnerabilities.");
+    } else {
+      logger.info("No new vulnerabilities were found.");
```
aberman, line 101: You should know that a customer will only see `info` if they run `gradle` with the `--info` flag.  Is that your intention?  Just how you can only see debug statements with the `--debug` flag.  If you want something that will print out by default, use `logger.lifecycle`

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      final String orgUuid,
+      final String serverName) {
+    ServerFilterForm serverFilterForm = new ServerFilterForm();
+    serverFilterForm.setApplicationIds(Collections.singletonList(applicationId));
+
+    Servers servers;
+    List<String> serverIds;
+
+    try {
+      serverFilterForm.setQ(URLEncoder.encode(serverName, "UTF-8"));
```
aberman, line 126: What in the world is `Q`?

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
```diff
@@ -1,15 +1,229 @@
...
+      servers = sdk.getServersWithFilter(orgUuid, serverFilterForm);
+    } catch (IOException e) {
+      throw new GradleException("Unable to retrieve the servers.", e);
+    } catch (UnauthorizedException e) {
+      throw new GradleException("Unable to connect to Contrast.", e);
+    }
+
+    if (!servers.getServers().isEmpty()) {
+      serverIds = new ArrayList<String>();
+      for (Server server : servers.getServers()) {
```
aberman, line 136: Could clean this up using `forEach`

aberman (approved):
Approved pending comments

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
BrianPhillips2020: yeah I had copied this over from the maven plugin, I think the sdk actually has a method for this

BrianPhillips2020: oh wait this is so stupid. The maven plugin, the sdk, and the old gradle plugin all have their own `SEVERITIES` list of Strings and their own form of this method, even though the sdk has an actual enum. The utility method I mentioned here returns `List<String>` even thought the ONLY place it's really needed is when it needs `List<RuleSeverity>` this project is giving me a headache

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
BrianPhillips2020: ah, yeah this is what I was unsure about, thank you

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastVerifyTestTask.java
BrianPhillips2020: I really do not know, the maven plugin adds it, the old gradle plugin added it, It doesn't appear to do anything on the `ServerFilterForm` in the sdk. Presumably it does....something....but I haven't found what yet

PR #129, opened 2025-03-25
JAVA-9127 update task registration to never be up to date

BrianPhillips2020:
Verify Test Tasks were being marked as `UP-TO-DATE` and not running if they passed once. Even running `clean` wouldn't cause them to rerun. Fixed by forcing this task to never be up to date. This is good anyway because we want to re-query the sdk every time we run tests or change the configuration anyway.

### DIFF
--- a/gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
+++ b/gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
@@ -106,6 +106,9 @@ public class ContrastGradlePlugin implements Plugin<Project> {
                                 sdk);
                     verifyTask.configure(
                         v -> {
+
+                          // ensure task outputs are never up to date so the task always runs
+                          v.getOutputs().upToDateWhen(task -> false);
                           v.dependsOn(testTask);
 
                           // set output file for found vulnerabilities
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -6,7 +6,7 @@ plugins {
     id 'com.gradle.plugin-publish' version '1.3.1'
 }
 
-version = '3.0.0'
+version = '3.0.1'
 group = 'com.contrastsecurity.gradle.plugin'
 
 repositories {


gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
```diff
@@ -106,6 +106,9 @@ public void apply(final Project target) {
                                 sdk);
                     verifyTask.configure(
                         v -> {
+
+                          // ensure task outputs are never up to date so the task always runs
+                          v.getOutputs().upToDateWhen(task -> false);
```
aberman, line 111: Weird API.  If it works, great, but the accepted way to do it is to use a Finalizer task: https://docs.gradle.org/current/userguide/controlling_task_execution.html#sec:finalizer_tasks 

> Finalizer tasks are executed even if the finalized task fails or if the finalized task is considered UP-TO-DATE

Wondering if you could do:

```
testTask.configure {
   finalizedBy verifyTask
}
```

No big deal either way.  Not sure when they added the API call you're using, you may want to verify that.

aberman (approved):

gradle-plugin/src/main/java/com/contrastsecurity/gradle/plugin/ContrastGradlePlugin.java
aberman: Good job finding that API, btw!  Gradle is a nest of unexplained APIs.

aberman: Actually, `finalizedBy` won't work in your case, if someone runs the verify task, we need the test task to run.

BrianPhillips2020: There actually isn't any situation where someone would run just the verify task by themselves, these are basically anonymous tasks that are registered by the plugin, but we wouldn't use `finalizedBy` for two reasons.
1. If the testTask is up to date we still want to run the verification, and if we finalize by the verifyTask, I'm pretty sure it would skip the finalized block if the test wasn't run.
2. We don't want the tests to always be verified when they run, only if `contrastCheck` is run.

BrianPhillips2020: > Good job finding that API, btw! Gradle is a nest of unexplained APIs.

thanks, what took longer was figuring out why and when the tasks weren't running.  I got lucky the first thing I tried just worked

PR #131, opened 2025-04-01
JAVA-8421 update maven release version

BrianPhillips2020:
I think the only thing we need to do to get the release to work is keep this `SNAPSHOT` version here. I'm going to try to release the plugin again. Worst case scenario it fails, best case scenario the existing build stuff auto strips the `SNAPSHOT` and updates the version on its own, which is I think what it's supposed to do. 🤞


tylernhoward (approved):

tylernhoward:
lets try it!

PR #137, opened 2025-04-11
JAVA-9137 update release workflow

BrianPhillips2020:
update maven release workflow. Will release from this branch once this is merged


maven-plugin/release-settings.xml
```diff
@@ -0,0 +1,32 @@
...
+
+    <profiles>
+        <profile>
+            <id>ossrh</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <gpg.passphrase>${env.MAVEN_GPG_PASSPHRASE}</gpg.passphrase>
+                <gpg.gpgArguments>--pinentry-mode,loopback</gpg.gpgArguments>
```
stiemannkj1, line 23: Add `--verbose` here too.

stiemannkj1: Actually maybe we should remove `--verbose` from both in case it leaks info.

.github/workflows/publish-maven-plugin.yml
```diff
@@ -38,22 +47,33 @@ jobs:
           git config --global user.name 'github-actions[bot]'
           git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
 
+
+      # Manually import gpg key so we don't have to deal with environment variable ambiguity.
+      - name: Import GPG Key 
+        run: |
+          rm -rf $HOME/.gnupg/
```
stiemannkj1, line 54: We should also probably clean up at the end of  the pipeline with an `always()` step. 🤷 prevents the key from being re-used in other pipelines when we don't want it to.

stiemannkj1 (approved):

maven-plugin/release-settings.xml
BrianPhillips2020: actually yeah that's probably a good idea

PR #159, opened 2026-06-08
Contrast Graph API

johnament:
## Summary

- Adds `ContrastGraphApi` interface and `ContrastGraphApiImpl` for the 5 v2 Contrast Graph endpoints served by `adr-explorer-aggregator` (graph search, incident graph, facets, application libraries, library details)
- Generates 17 model POJOs from a self-contained `contrast-graph.yaml` OpenAPI spec via `openapi-generator-maven-plugin` (models only, `okhttp-gson` library, `dateLibrary=string`)
- Exposes `ContrastSDK.graphApi()` factory method parallel to the existing `scan()` factory
- Adds `JSON.java` utility (delegates to `GsonFactory`) required by generated model convenience methods

## Notable build changes

- `openapi-generator-maven-plugin 7.9.0` added (models-only, no API stubs)
- Lombok bumped `1.18.18 → 1.18.30` for JDK 21 annotation processor compatibility
- `jsr305` and `javax.annotation-api` added as optional compile-scope deps (needed by generated `@Nullable` / `@Generated` annotations, not needed at runtime)

## Test plan

- [ ] `ContrastGraphApiFactoryTest` — verifies `ContrastSDK.graphApi()` returns a `ContrastGraphApiImpl` instance
- [ ] `ContrastGraphApiImplPactTest` — 5 Pact consumer contract tests, one per API method, all asserting specific field values
- [ ] Pre-existing `CodeArtifactsImplTest` / `ProjectsImplTest` / `ScansImplTest` failures are a Mockito 3.11.2 + JDK 21 incompatibility that predates this PR — not introduced here

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tylernhoward (approved):

sdk/src/main/java/com/contrastsecurity/sdk/JSON.java
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.contrastsecurity.sdk;
+
+/*-
```
tylernhoward, line 1: This seems like the license we would want above instead, accidental inclusion?

sdk/src/test/java/com/contrastsecurity/sdk/graph/ContrastGraphApiImplPactTest.java
```diff
@@ -0,0 +1,283 @@
...
+  private static final String LIBRARY_HASH =
+      "3a42c503953909637f78dd8c99b3b85ddde362415585afc11901bdefe8349102";
+  private static final String AGENT_INSTANCE_ID = "abc1234";
+
+  private ContrastGraphApiImpl client(final MockServer server) {
+    ContrastSDK sdk =
+        new ContrastSDK.Builder("user", "serviceKey", "apiKey")
+            .withApiUrl(server.getUrl() + "/Contrast/api")
+            .build();
+    return new ContrastGraphApiImpl(sdk, new Gson());
```
tylernhoward: Should we use `JSON.getJson()` here instead?

sdk/src/main/java/com/contrastsecurity/sdk/graph/ContrastGraphApiImpl.java
tylernhoward, line 1: nit: possibility to use a helper method to reuse code in this class.

BrianPhillips2020 (approved):

sdk/src/main/java/com/contrastsecurity/sdk/graph/ContrastGraphApiImpl.java
tylernhoward: I think I'm ok with this as is, I think any refactor to not duplicate code affects readability

johnament: I did end up cleaning up the calls as it was ignoring proxy settings the way written.

